PR #4, opened 2023-06-14
Generate RELEASE file

ericonr:


Dockerfile
mmdonatti, line 1: It wouldn't be better to link BUILD_STAGE_VERSION with image releases?

ericonr: I think it would be ideal, but I didn't do it because this was still experimental.

And now the source I used for sequencer is down, because of Bessy's cyberattack, so I can't tag and upload an image created with CI. Will open an issue about it, though.

PR #7, opened 2023-07-11
Add support for areaDetector

henriquesimoes:
I've added support for the core area detector modules and `ADSimDetector` for now. My intention is to add other modules, which can be done by generating a `RELEASE` file with the corresponding module uncommented and installing its EPICS and system dependencies.

Suggestions and feedbacks are welcome!


base/install_area_detector.sh
```diff
@@ -0,0 +1,78 @@
...
+$module_releases
+
+ASYN=${EPICS_MODULES_PATH}/asyn
+" > RELEASE.local
+
+ln -s RELEASE.local RELEASE_PRODS.local
+ln -s RELEASE.local RELEASE_LIBS.local
+
+echo "
+BUILD_IOCS=NO
```
ericonr, line 40: Why do this instead of appending to the existing file?

I can probably be convinced that it's worth it so we have to properly inspect changes between versions instead of doing the least effort, but I'd like to know if that's the goal :)

henriquesimoes: Yeah, that's the point. If anything changes in this configuration file in the `areaDetector` repo, I'd like to have it explicitly and properly handled and well documented in this repository.

base/install_area_detector.sh
```diff
@@ -0,0 +1,77 @@
...
+
+WITH_BOOST=NO
+WITH_PVA=NO
+WITH_QSRV=NO
```
ericonr, line 48: Why not build the QSRV plugin, if we are already here?

base/install_area_detector.sh
```diff
@@ -0,0 +1,77 @@
...
+
+WITH_NEXUS=YES
+NEXUS_EXTERNAL=NO
+
+WITH_OPENCV=NO
+
+WITH_SZIP=YES
+SZIP_EXTERNAL=NO
+
+XML2_EXTERNAL=NO
```
ericonr, line 76: On your machine, when you showed me this, you were installing the XML2 package, no?

ericonr (commented):
In commit messages, "add support *for*", instead of "*to*".

I'm also pending towards calling it areaDetector in commit messages, since it's the official name... I know we talked about this earlier.

base/install_area_detector.sh
henriquesimoes: Only because I don't know any easy way to test if it is working properly yet. PV Access is still completely new for me. Should we install anything else to make it work or should it be just enable it in this configuration?

base/install_area_detector.sh
henriquesimoes: Yes, as it is Aravis library dependency. `ADCore` by default uses the `ADSupport` version of `libxml2`, and that's why it wasn't changed there. But I had configured this before our in-person discussion.

I may change it to use the externally installed library. But before that I'd like to confirm if you agree based on some new things I've found out. Basically, `ADSupport`

- [uses version 2.9.4](https://github.com/areaDetector/ADSupport/blob/ac5d261a1abcddba11827a0b5ce500437e4485c9/supportApp/xml2Src/README.epics#L1) while Debian has packaged [version 2.9.10 for Debian 11.7](https://packages.debian.org/bullseye/libxml2) (and [2.9.14 for Debian 12](https://packages.debian.org/bookworm/libxml2)). Memory leaks and other bugs are fixed after 2.9.4, but I'm not sure how critical they are.
- patches [things for operating systems other than Linux](https://github.com/areaDetector/ADSupport/blob/master/supportApp/xml2Src/README.epics). These patches do not break anything from the original API.
- builds `libxml2` without ICU, HTML, FTP and `iconv` support, and thus removes the need to have `libicu67` in the system, whose version 67.1-7 has installation size of 33.9MB (as we checked before).
   
   [ICU support is indeed not required in Linux in `libxml` 2.9.4](https://gitlab.gnome.org/GNOME/libxml2/-/blob/v2.9.4/INSTALL.libxml2?ref_type=tags#L36) and latest versions. It has been [included as a dependency in Debian due to "longstanding issues with chromium"](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=776254), and there is already [a bug report for removing it](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=962464) due to the very same reason we are not willing to add it to our IOC containers by default.

Based on that, we could use the `ADSupport` one to avoid the cumbersome ICU depencency (also making it visible to the dynamic linker) or use the packaged version with the external installation being specified both in `BUILD_PACKAGES` and `RUNTIME_PACKAGES` by each `areaDetector` IOC as we discussed previously.

In the long-term, I see it is better to have it externally installed (considering [Debian Bug 962464](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=962464) is fixed). I'm not so sure in the sort term, though.

base/install_area_detector.sh
ericonr: I'd like for this rationale to be documented in the commit message, then.

base/install_area_detector.sh
ericonr: Building with qsrv only allows us to use the pvaccess plugin, which is opt-in. I'd rather have it built than not.

base/install_area_detector.sh
ericonr: Why do we need to specify it in `BUILD_PACKAGES`? The build image is huge already, so I think it could be there by default.

The 30MB difference is relevant when building the small images for static IOCs, but it`s not as relevant when we consider the size (at least for now) of the dynamic IOC images.

Furthermore, I thoroughly dislike the idea of having two versions of the same library being used in the same final executable. It`s risky and, in this case, doesn't bring a lot of benefits. I think that's the main reason I'm in favor of using the system xml2 library.

henriquesimoes: > Why do we need to specify it in BUILD_PACKAGES? The build image is huge already, so I think it could be there by default.

My mistake. I meant including it the the `base/Dockerfile` actually. Otherwise, it won't even build `ADCore`.

>  it`s not as relevant when we consider the size (at least for now) of the dynamic IOC images.

You have a good point. I agree. Cleaning up other things may be more effective for now.

> Furthermore, I thoroughly dislike the idea of having two versions of the same library being used in the same final executable.

Yeah, having both looks messy. I was talking about having one or the other, really. But as we talked in person, having it installed in the system seems the cleaner way to have the other dependent software working, such as `libaravis`.

ericonr (commented):
Exporting `DOCKER_BUILDKIT=1` when building the IOC image can be necessary on some systems (it was on mine).

Commit message fix suggestion:

> This new stage makes it possible to dynamically link to EPICS and other
system libraries, including areaDetector support libs. This solves the
issue of finding a way to pass the information to the EPICS build system
on how to link all necessary libraries dynamically while the EPICS libraries remain statically
linked.

> This makes it easier to find compatible versions for all submodules, since they are defined in the supermodule.

> Configuration files are written from scratch, since this makes the script
cleaner and less coupled with the upstream configuration. In addition,
any change in the upstream repository needs to be explicitly and
properly handled here.

> `RELEASE_PRODS.local` and `RELEASE_LIBS.local` mirror `RELEASE.local`
configuration, as release files can, in theory, define extra modules
paths without affecting the build. This is not true only for modules
that are compiled conditionally based on their path being defined. It
is not expected to be the case here.

> `CONFIG_SITE.local` has been configured so that we can track down
changes in configuration more easily.

> Several libraries from ADSupport were configured to be built (as
it is the default value). MagicGraphicK is only required by ADFileMagick
and requires some system libraries, so won't be handled for now.

> libxml2` is installed from the Debian package.

> and it avoids having two versions of the same library linked into a single
executable (which can lead to undesired side-effects) and customized
dynamic loader configuration.

> In order for it to work with `libnanohttp_stream` from
`ADSupport/supportApp/xml2Src`, a patch is required to fix the build; it should be removed once accepted upstream.

(the context is in the patch file, doesn't need to be here)

> since only detector IOCs based on ADAravis will need it at runtime.

> While it is not restored, we are hosting the
needed files in an alternative server. This should be reverted once the official servers back online.

README.md
```diff
@@ -39,6 +40,11 @@ services:
         ENTRYPOINT: ./YOUR_ENTRYPOINT
 ```
 
+By default, the IOC will be built with statically linked libraries. If you want
+to use dynamically linked libraries, you should define the build target as
```
ericonr, line 44: "EPICS libraries" instead of "libraries".

It's "if you *need* to use", not "want" (use bold, even)

README.md
```diff
@@ -39,6 +40,11 @@ services:
...
+`dynamic-link`. This will increase the resulting image size, since unneeded
+dependencies will also be copied.
```
ericonr, line 46: "This will increase the resulting image size, since unused dependencies will also be copied."

base/install_area_detector.sh
```diff
@@ -66,13 +66,17 @@ WITH_OPENCV=NO
 WITH_SZIP=YES
 SZIP_EXTERNAL=NO
 
-XML2_EXTERNAL=NO
+XML2_EXTERNAL=YES
+XML2_LIB=/usr/lib/x86_64-linux-gnu
```
ericonr, line 78: I don't love that this hardcodes x86_64... Especially in light of us having talked to Marco about using this for mobipix.

I think you should be able to get something using `pkg-config` to print `libdir`

ericonr:
Feel free to add a commit to use `v0.2.0git` in the ioc Dockerfile, if you'd like, btw.

base/install_area_detector.sh
henriquesimoes: At first, I didn't think libxml2 had `pkg-config` support. It turns out I mistook its name.

I've used `pkg-config` to determine the `XML2_INCLUDE` path as well. Same approach is used for GLib and Aravis.

README.md
```diff
@@ -39,6 +40,16 @@ services:
...
 
+By default, the IOC will be built with statically linked EPICS libraries. If
+you **need** to link them dynamically, you must define the build target as
+`dynamic-link`. This will increase the resulting image size, since unused
+dependencies will also be copied.
+
+Enabling [BuildKit](https://docs.docker.com/build/buildkit/) (for instance, by
+exporting `DOCKER_BUILDKIT=1`) when building the IOC image can be necessary on
+some systems, because the classic builder goes through all stages even when
+they are not need and they fail.
```
ericonr, line 51: I think this wording is a bit confusing... I'd say something like:

```
Some Docker versions don't use [BuildKit](https://docs.docker.com/build/buildkit/) by
default, and it is necessary to enable it by exporting `DOCKER_BUILDKIT=1` when
building the IOC image, because the classic builder goes through all stages even
when they are not needed or are known to fail.
```

henriquesimoes: I understand. I was willing to let it up to the user define how she want to enable BuildKit. It may also be enabled through Docker daemon's configuration file, which can be more practical. But I agree the wording can be improved. How about something like this?

> Some Docker versions don't use [BuildKit](https://docs.docker.com/build/buildkit/) by default, and it is necessary to enable it, for instance, by exporting `DOCKER_BUILDKIT=1` when building the IOC image, because the classic builder goes through all stages even when they are not needed or are known to fail.

ericonr: I hadn't thought about changing the daemon's default :) That's a good point.

ericonr (approved):

PR #9, opened 2023-08-11
ioc: reorder stages to improve portability.

henriquesimoes:
Docker without BuildKit and Podman fail when some unused stage fail before the defined target. As static build stage from the IOC image is known to possibly fail, we now define all static build stages last.

Arguments have now been define only in stages they are used. This is required to support non-BuildKit builds, as they aren't inherited.


Dockerfile
```diff
@@ -1,11 +1,9 @@
-ARG BUILD_STAGE_VERSION=v0.2.0
+ARG BUILD_STAGE_VERSION=latest
```
ericonr, line 1: This change slipped through.

ericonr (changes requested):
In commit message:

> Docker without BuildKit and Podman fail when some unused stage *fails*

> Arguments have now been *defined* only in stages they are used.

Dockerfile
henriquesimoes: Oops. Fixed!

ericonr (approved):

PR #10, opened 2023-08-15
base: include libtiff in areaDetector build

henriquesimoes:
I've included `libtiff` in the build so that we can use [`NDFileTIFF`](https://areadetector.github.io/master/ADCore/NDFileTIFF.html) plugin.

I've mistakenly left it off from #7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,8 +63,8 @@ registry.
 ### areaDetector IOCs
 
 `areaDetector` IOCs must be built with target `dynamic-link`. In addition, they
-must include `libxml2` in the `RUNTIME_PACKAGES`, as it is not built in
-`ADSupport`.
+must include `libxml2` and `libtiff5` in the `RUNTIME_PACKAGES`, as they are
+not built in `ADSupport`.
 
 ### Possible issues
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -63,8 +63,8 @@ registry.
 ### areaDetector IOCs
 
 `areaDetector` IOCs must be built with target `dynamic-link`. In addition, they
-must include `libxml2` in the `RUNTIME_PACKAGES`, as it is not built in
-`ADSupport`.
+must include `libxml2` and `libtiff5` in the `RUNTIME_PACKAGES`, as they are
+not built in `ADSupport`.
```
ericonr, line 67: You use `libtiff-4` in `pkg-config`, but `libtiff5` in the package name. Is that correct? If so, the commit should mention the discrepancy.

henriquesimoes: Yes, it is (unfortunately) correct. [`libtiff5` (and `libtiff-dev`) sources libtiff 4.2.0](https://packages.debian.org/source/bullseye/tiff), hence `libtiff-4` for `pkg-config`. I don't know why the package name was bumped, though.

But I agree it's better for our sanity to make this explicit in the message.

ericonr: Thanks!

ericonr (approved):

PR #11, opened 2023-08-18
Added motor and ipac modules

guirodrigueslima:


base/install_motor.sh
```diff
@@ -0,0 +1,41 @@
...
+echo "$module_releases" >> ${EPICS_MODULES_PATH}/../RELEASE
+
+cd ../configure
+
+echo "
+EPICS_BASE=${EPICS_BASE_PATH}
+MOTOR=${EPICS_MODULES_PATH}/motor
+ASYN=${EPICS_MODULES_PATH}/asyn
+SNCSEQ=${EPICS_MODULES_PATH}/seq
+AUTOSAVE=${EPICS_MODULES_PATH}/autosave
```
henriquesimoes, line 26: Do you really need autosave for building the support module?

base/install_motor.sh
```diff
@@ -0,0 +1,41 @@
...
+BUSY=${EPICS_MODULES_PATH}/busy
+IPAC=${EPICS_MODULES_PATH}/ipac
+" > RELEASE.local
+
+ln -s RELEASE.local RELEASE_PRODS.local
+ln -s RELEASE.local RELEASE_LIBS.local
```
henriquesimoes, line 32: This is a very areaDetector specific way of configuring things. It seems that [Motor](https://github.com/epics-modules/motor/tree/d1ed86f128733070ec720286d1e1fad05e8e8156/configure) does not need to define neither `RELEASE_PRODS.local` nor `RELEASE_LIBS.local`.

(TBH, I think even areaDetector should not need them. That's why I have symlinked them.)

base/install_motor.sh
```diff
@@ -0,0 +1,41 @@
...
+
+echo "
+BUILD_IOCS=YES
```
henriquesimoes, line 35: I recall you telling me we don't use IOCs directly from the support motor modules. Is that right? If so, I'd rather have this set to `NO` instead. We don't need to build unused software.

base/install_motor.sh
```diff
@@ -0,0 +1,41 @@
...
+git clone --depth 1 --branch ${MOTOR_VERSION} \
+    https://github.com/epics-modules/motor
+
+cd motor/modules
+
+git submodule update --init --depth 1 -j ${JOBS} \
+    motorMotorSim
+
+module_releases="
+MOTOR=${EPICS_MODULES_PATH}/motor
```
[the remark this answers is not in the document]
guirodrigueslima: All motor IOCs need the $MOTOR reference to load the default motor record .db files.

henriquesimoes (changes requested):
Commit message can (and should) be more descriptive. I'd expect something like:

```
base: add support for motor.

Similarly to areaDetector, a separate script has been used to configure motor
module and its submodules due to its different setup.

IOCs in the support module are not built, since we currently do not use any of
them exactly the same way as they are configured in the community repositories.
This prevents the image size from increasing with unused build artifacts.

IPAC module is built to support `motorHytec`. This way, all motor modules
provided by the motor (super)module are supported.
```

ericonr (approved):
Pending @henriquesimoes 's final review, I'm okay with merging this.

Thanks for the great work!

henriquesimoes (approved):
It's fine by me as well. Thanks for working on this, @guirodrigueslima.

PR #12, opened 2023-08-29
Fix CI workflow for forked repositories

henriquesimoes:
I've changed the CI so that it works on forked repositories by using GItHub context variables. Although there are more build environment variables now, we can still build the image locally only defining the `TAG` (and `JOBS`) variables.

I've also fixed the issue of having tag pushes ignored. It works now. I've already tested in my fork. However, I'm not sure if you would like to keep with the idea of having any tag triggering the CI.

### DIFF
--- a/.github/workflows/base-image.yml
+++ b/.github/workflows/base-image.yml
@@ -25,13 +25,13 @@ jobs:
       - name: Build image
         run: cd base && JOBS=$(nproc) docker compose build
       - name: Log in to registry
-        if: github.event_name == 'push'
+        if: github.event_name == 'push' && vars.PUSH_TO_REGISTRY == 'true'
         run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u $ --password-stdin
       - name: Push tagged image
-        if: github.event_name == 'push' && github.ref_type == 'tag'
+        if: github.event_name == 'push' && vars.PUSH_TO_REGISTRY == 'true' && github.ref_type == 'tag'
         run: docker push $REGISTRY/$IMAGE:$TAG
       - name: Push latest image
-        if: github.event_name == 'push' && github.ref_type == 'branch'
+        if: github.event_name == 'push' && vars.PUSH_TO_REGISTRY == 'true' && github.ref_type == 'branch'
         run: |
           docker tag $REGISTRY/$IMAGE:$TAG $REGISTRY/$IMAGE:latest
           docker push $REGISTRY/$IMAGE:latest
--- a/.github/workflows/base-image.yml
+++ b/.github/workflows/base-image.yml
@@ -1,7 +1,8 @@
 name: Base image build
 on:
   push:
-    tags-ignore:
+    tags:
+      - 'v*'
     branches:
       - main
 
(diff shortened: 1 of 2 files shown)


.github/workflows/base-image.yml
```diff
@@ -26,7 +29,9 @@ jobs:
...
       - name: Push tagged image
         if: github.event_name == 'push' && github.ref_type == 'tag'
-        run: docker push ghcr.io/cnpem/lnls-debian-11-epics-7:$TAG
+        run: docker push $REGISTRY/$IMAGE:$TAG
       - name: Push latest image
         if: github.event_name == 'push' && github.ref_type == 'branch'
-        run: docker tag ghcr.io/cnpem/lnls-debian-11-epics-7:$TAG ghcr.io/cnpem/lnls-debian-11-epics-7:latest && docker push ghcr.io/cnpem/lnls-debian-11-epics-7:latest
+        run: |
+          docker tag $REGISTRY/$IMAGE:$TAG $REGISTRY/$IMAGE:latest
+          docker push $REGISTRY/$IMAGE:latest
```
ericonr, line 38: This might be going in the opposite direction of what you intended, but I think the correct behavior is actually not publishing anything if we aren't building in the main repository :thinking: 

It'd be a big surprise to fork a repo and suddenly you're filling up your personal registry...

Therefore, I'd consider adding another condition to the if for this block, where the repository has to be in our org.

.github/workflows/base-image.yml
```diff
@@ -9,6 +9,9 @@ on:
 
 env:
   TAG: ${{ github.event_name == 'push' && github.ref_name || github.head_ref }}
+  REGISTRY: ghcr.io/${{ github.repository_owner }}
+  IMAGE: lnls-debian-11-epics-7
+  SOURCE: https://github.com/${{ github.repository }}
```
ericonr, line 15: That said, defining these variables externally is a good idea.

.github/workflows/base-image.yml
```diff
@@ -1,7 +1,8 @@
 name: Base image build
 on:
   push:
-    tags-ignore:
+    tags:
+      - '**'
```
ericonr, line 5: Since we are using `v*` versions, be explicit about only including them here.

ericonr (changes requested):

.github/workflows/base-image.yml
henriquesimoes: Okay, I haven't considered that it will fill up one's registry even when they only want to contribute upstream.

I'll fix that up. Thanks for the feedback!

ericonr (commented):
For commit messages:

> All constant values related to this repository *have* been defined as
variables defaulting to our configuration.

> Image name has also been extracted to a constant variable, so that we
don't need to update all its occurrences manually*, which* could be error
prone.

> Having the *push event trigger a container image push by default* for forked
repositories would fill up one's registry without any deliberate action
regarding this.

> *Repository owners* need to set
the PUSH_TO_REGISTRY GitHub Actions variable to 'true' to enable the push
to registry *for any of the events* [1].

ericonr (approved):

PR #13, opened 2023-08-29
base: use official Sequencer package source.

henriquesimoes:
Official servers are [(in theory) back online][1] for serving Sequencer package. But now with a new domain name.

[1]: https://epics.anl.gov/tech-talk/2023/msg01214.php


henriquesimoes:
Even though they are supposedly back online, it seems to be answering 404 for [Sequencer 2.2.9](https://epics-sequencer.sourceforge.io/releases/seq-2.2.9.tar.gz). Have you heard any news about it?

ericonr:
It seems to be the case for all releases. I will throw them an email, because it hasn't been mentioned there yet.

ericonr (approved):

PR #14, opened 2023-08-31
ioc: remove .git directory during build.

ericonr:
The .git directory for submodules isn't being removed by this command, but we will assume those are smaller, so not worth the effort to remove cleanly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,7 @@ WORKDIR /opt/${REPONAME}
 COPY . .
 
 RUN cp /opt/epics/RELEASE configure/RELEASE
+RUN rm -rf .git/
 
 
 FROM debian:${DEBIAN_VERSION}-slim as base


Dockerfile
```diff
@@ -13,6 +13,7 @@ WORKDIR /opt/${REPONAME}
 COPY . .
 
 RUN cp /opt/epics/RELEASE configure/RELEASE
+RUN rm -rf .git/
```
henriquesimoes, line 16: This adds another caching layer, which shouldn't be needed. It may slow down the build a tiny bit.

Is there any special reason you made this another `RUN` layer?

henriquesimoes (commented):
Doesn't `.git/modules` store the `.git` for submodules?

ericonr:
> Doesn't .git/modules store the .git for submodules?

Oh wow, I didn't know this. Hadn't checked that the `.git` inside submodules is just a file.

Dockerfile
ericonr: Mostly because the `static-build` stage has separate RUNs for commands which weren't as closely related.
I'd be okay with moving this into the RUN before it.

henriquesimoes (commented):
Please, update the commit message, since it does remove submodules .git as well. :)

henriquesimoes (approved):

PR #15, opened 2023-09-01
Add support for Pimega detectors

henriquesimoes:
These changes are using locally served packages, since nothing is available in the Internet.


ericonr (commented):
Nothing release breaking here

Dockerfile
```diff
@@ -38,6 +40,11 @@ RUN ln -s ${ENTRYPOINT} ./entrypoint
 ENTRYPOINT ["./entrypoint"]
 
 
+FROM base AS no-build
+
+COPY --from=build-image /opt /opt
```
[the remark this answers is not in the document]
ericonr: Seems reasonable.

Dockerfile
```diff
@@ -1,7 +1,9 @@
 ARG BUILD_STAGE_VERSION=v0.2.0
 ARG DEBIAN_VERSION=11.7
 
-FROM ghcr.io/cnpem/lnls-debian-11-epics-7:${BUILD_STAGE_VERSION} AS build-stage
+FROM ghcr.io/cnpem/lnls-debian-11-epics-7:${BUILD_STAGE_VERSION} AS build-image
+
+FROM build-image as build-stage
```
ericonr, line 6: `AS` seems to be the standard for our `FROM` directives.

Dockerfile
```diff
@@ -38,6 +40,11 @@ RUN ln -s ${ENTRYPOINT} ./entrypoint
 ENTRYPOINT ["./entrypoint"]
 
 
+FROM base AS no-build
+
+COPY --from=build-image /opt /opt
+
+
```
ericonr, line 57: Shouldn't this be above `build-stage`, even? Otherwise, the build-stage will fail for bad resolvers.

Dockerfile
```diff
@@ -22,17 +22,27 @@ FROM debian:${DEBIAN_VERSION}-slim as base
...
     rm -rf /var/lib/apt/lists/*
 
+RUN for url in $RUNTIME_TAR_PACKAGES; do \
+        download_dir=$(mktemp -d) && \
+        wget $url -P $download_dir && \
+        tar --no-same-owner -xvf $download_dir/* -C /; \
+    done && \
+    ldconfig && \
+    rm -rf /tmp/*
+
```
ericonr, line 45: As we discussed, this should be a separate patch (and it was going to use our tar helper, no?)

Also, `--no-same-owner` should be explained quickly in the commit.

base/Dockerfile
```diff
@@ -19,6 +21,15 @@ RUN apt update -y && \
         wget \
         ca-certificates
 
+RUN TAR_PACKAGES="http://gca-jobs:1234/packages/libpimega_2.5.1-0_amd64.tar.gz" && \
```
ericonr: The libpimega version should be a variable.

base/Dockerfile
```diff
@@ -21,7 +21,9 @@ RUN apt update -y && \
         wget \
         ca-certificates
 
-RUN TAR_PACKAGES="http://gca-jobs:1234/packages/libpimega_2.5.1-0_amd64.tar.gz" && \
+RUN TAR_PACKAGES="\
+        http://gca-jobs:1234/packages/libpimega_2.5.1-0_amd64.tar.gz \
+        http://gca-jobs:1234/packages/libsscpimega_0.8.12-0_amd64.tar.gz" && \
```
ericonr, line 26: Version should also be a variable. And libsscpimega should be built from source.

Dockerfile
henriquesimoes: Good catch. My mistake.

base/Dockerfile
henriquesimoes: It'll be built from source as soon as it becomes open source. :)

I'll include hard-coded versions in `.env` instead.

ericonr: We can still host a tarball at gca-jobs with the source. That way, we can deal with the necessary build (and deployment - if it's installed in `/usr/local`, it won't be copied automatically, so installing in `/opt` might make sense) steps before it's hosted on github.

Dockerfile
henriquesimoes: Yes, you're right. I should pay attention to this and test with podman more frequently. Thanks!

base/Dockerfile
henriquesimoes: That's a good point. I may work on this build and installation process while still using the tarball. I may think more about it ans sketch somethings after working on other pendencies, such as script for installing tarballs and extracting the PRs for tarball install and `no-build` separately.

base/Dockerfile
henriquesimoes: `NDSSCPimega` should also follow this logic, right?

ericonr: Yup

PR #20, opened 2023-09-13
ioc: add stage without IOC building.

henriquesimoes:
This allows us to run an IOC which was built along with its support module. This pattern is used for some community modules, including motor's and areaDetector's ones.

A new (empty) stage has been created only to specify the base image in the COPY command, since `--from` argument [does not support variable expansion][1].

`build-stage` has been moved below `no-build` stage, so bad multi-stage resolvers won't fail for the new stage.

[1]: https://forums.docker.com/t/how-can-i-expand-a-variable-within-a-copy-command-in-the-dockerfile/113538/2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,8 @@
 ARG DEBIAN_VERSION=11.7
 
-FROM ghcr.io/cnpem/lnls-debian-11-epics-7:v0.3.0 AS build-stage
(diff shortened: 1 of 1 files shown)


ericonr (approved):
Assuming the build has been tested with podman/non buildkit docker already, approved.

PR #21, opened 2023-09-15
Install tarball packages with custom extraction

henriquesimoes:
Installing through tarballs is useful to enable us to include software that is not yet packaged in the distribution repositories.

A custom extraction script is interesting to solve, in a centralized way, permission issues in rootless container engines both for source code and binaries packages.

I've named the custom script `untar`, but I'm open to suggestions for other names.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,7 @@ FROM debian:${DEBIAN_VERSION}-slim AS base
 ARG RUNDIR
 ARG ENTRYPOINT=/bin/bash
 ARG RUNTIME_PACKAGES
+ARG RUNTIME_TAR_PACKAGES
 
 RUN apt update -y && \
     apt install -y --no-install-recommends \
@@ -14,9 +15,14 @@ RUN apt update -y && \
         busybox \
         netcat-openbsd \
         procserv \
+        wget \
         $RUNTIME_PACKAGES && \
     apt clean && \
     rm -rf /var/lib/apt/lists/*
+
+COPY --from=build-image /usr/local/bin/lnls-get-n-unpack /usr/local/bin/lnls-get-n-unpack
+RUN lnls-get-n-unpack -r $RUNTIME_TAR_PACKAGES && \
+    ldconfig
 
 WORKDIR ${RUNDIR}
 
@@ -34,8 +40,10 @@ FROM build-image AS build-stage
 
 ARG REPONAME
 ARG BUILD_PACKAGES
+ARG BUILD_TAR_PACKAGES
 
 RUN if [ -n "$BUILD_PACKAGES" ]; then apt update && apt install $BUILD_PACKAGES; fi
+RUN lnls-get-n-unpack -r $BUILD_TAR_PACKAGES
 
 WORKDIR /opt/${REPONAME}
 
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -19,6 +19,8 @@ RUN apt update -y && \
         wget \
         ca-certificates
 
+COPY lnls-get-n-unpack.sh /usr/local/bin/lnls-get-n-unpack
+
 ARG EPICS_BASE_VERSION
 ENV EPICS_BASE_PATH /opt/epics/base
 ENV EPICS_MODULES_PATH /opt/epics/modules
--- a/base/lnls-get-n-unpack.sh
+++ b/base/lnls-get-n-unpack.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+#
+# Download and extract tarball archive from the network.
+
+set -eu
+
+case "$1" in
+    -r) dest=/ ;;
+    -l) dest=. ;;
+    *) >&2 echo "Invalid extraction mode: must be either root (-r) or local (-l)."
+       exit 1;
+    ;;
+esac
+
+shift
+
+for url; do
+    download_dir=$(mktemp -d)
+
+    echo Downloading "$url"...
+    wget -P $download_dir -o /tmp/wget.log "$url" || (cat /tmp/wget.log && false)
+    tar --no-same-owner -xf $download_dir/* -C $dest
+    rm -rf $download_dir /tmp/wget.log
+done
--- a/base/install_epics.sh
+++ b/base/install_epics.sh
@@ -2,9 +2,7 @@
 
 set -ex
 
-wget https://epics-controls.org/download/base/base-${EPICS_BASE_VERSION}.tar.gz
-tar -xf base-${EPICS_BASE_VERSION}.tar.gz
-rm base-${EPICS_BASE_VERSION}.tar.gz
+lnls-get-n-unpack -l https://epics-controls.org/download/base/base-${EPICS_BASE_VERSION}.tar.gz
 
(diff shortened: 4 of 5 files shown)


base/untar.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+#
+# Download and extract tarball archive from the network.
+
+set -ex
```
ericonr, line 5: `-x` left over. If you'd like to track progress, we should add progress messages, IMO.
Would it make sense to add `-u` here as well?

base/untar.sh
```diff
@@ -0,0 +1,26 @@
...
+    *) >&2 echo "Invalid extraction mode: must be either root (-r) or local (-l)."
+       exit 1;
+    ;;
+esac
+
+shift
+
+urls=$@
+
+for url in $urls; do
```
ericonr, line 19: This can be done as `for url; do`, which uses the `$@` arguments automatically

base/untar.sh
```diff
@@ -0,0 +1,26 @@
...
+    download_dir=$(mktemp -d)
+    wget $url -P $download_dir
+    tar --no-same-owner -xvf $download_dir/* -C $dest
+    rm -rf $download_dir
+done
+
+ldconfig
```
ericonr, line 26: This should only be done in places where we know libraries are being installed, after running the script.

base/Dockerfile
```diff
@@ -19,6 +19,8 @@ RUN apt update -y && \
         wget \
         ca-certificates
 
+COPY untar.sh /usr/local/bin/untar
```
ericonr, line 22: `untar` doesn't feel like it covers everything the script is doing. What do you think about something like `lnls-get-install` instead?

ericonr (changes requested):

base/untar.sh
henriquesimoes: I'd say it's a left over for the other scripts. Here I've only kept the pattern of the other scripts. Do you see any reason not to use `-x` here, but use it for `install_*.sh`?

About `-u`, considering that one wouldn't use the script if `$@` is empty, it might be interesting to include `-u` indeed.

base/untar.sh
henriquesimoes: Nice! I didn't know about that. Thanks!

base/Dockerfile
henriquesimoes: > `untar` doesn't feel like it covers everything the script is doing.

Indeed it misses the `wget` part. I had considered `wuntar`, but it starts to get weird.

> What do you think about something like `lnls-get-install` instead?

I've avoided using "install" because installing a source code is not a terminology I'm used to. If you think that's okay to call extracting an archive in the local directory as install, we may stick with it.

I dispensed with using "lnls" as it is just an internal utility (although it still resides on the IOC filesystem afterwards). We may add it if you think it helps distinguish the special purpose scripts we create.

base/untar.sh
ericonr: > I'd say it's a left over for the other scripts. Here I've only kept the pattern of the other scripts. Do you see any reason not to use -x here, but use it for install_*.sh?

At most that it would pollute the build output for the actual IOC images. But I hadn't considered that the other scripts had it already.

base/Dockerfile
ericonr: As we discussed, `lnls-` is interesting due to namespacing.

base/untar.sh
ericonr: Yeah, mostly just going through "downloading x..."

I'd be in favor of collapsing `tar -x` output, yeah.

henriquesimoes: I've redirected the output from `wget` as well. Let me know if there is anything we can improve about that (maybe keeping the entire URL?).

base/lnls-get-n-unpack.sh
```diff
@@ -0,0 +1,24 @@
...
+    ;;
+esac
+
+shift
+
+for url; do
+    download_dir=$(mktemp -d)
+
+    echo "Downloading $(basename $url)..."
+    wget $url -P $download_dir &> /dev/null
```
ericonr, line 21: I'd quote both instances of `$url` here, just to be sure :)

Dockerfile
```diff
@@ -7,17 +7,23 @@ FROM debian:${DEBIAN_VERSION}-slim AS base
...
         procserv \
+        wget \
         $RUNTIME_PACKAGES && \
     apt clean && \
     rm -rf /var/lib/apt/lists/*
 
+COPY --from=build-image /usr/local/bin/lnls-get-n-unpack /usr/local/bin/lnls-get-n-unpack
+RUN lnls-get-n-unpack -r $RUNTIME_TAR_PACKAGES && \
+    ldconfig
+
```
ericonr, line 26: This only makes sense if you are downloading the same/similar tarball while building base, or if RUNTIME_TAR_PACKAGES are installed at build time as well when you're building the final IOC... Do you agree? (mostly just wondering if for commit internal consistency, we should add something else here)

base/lnls-get-n-unpack.sh
ericonr: Do `wget -P $download_dir -o $download_dir/wget-log  "$url" || (cat $download_dir/wget-log && false)`

Dockerfile
henriquesimoes: Not necessarily, although I doubt it is the case for us now. A runtime script might depend on other utility (not packaged for Debian), which isn't required during the build (a vendor executable, for instance). Being possible to call system applications in IOC Shells suggests that this situation can occur.

What do you propose regarding this fact of probably always installing packages also used in other build phases?

base/lnls-get-n-unpack.sh
henriquesimoes: Good idea! Thanks for the suggestion.

Dockerfile
ericonr: That the build-stage preparation for build should maybe include a similar RUN (without the ldconfig, though?)

I see your point and could leave this commit as is, too.

henriquesimoes: Well, you have a good point. When a library is not built in the base image, it will require a way to install during the `build-stage` as well. I've added it now. Thanks for the suggestion.

base/lnls-get-n-unpack.sh
```diff
@@ -0,0 +1,24 @@
...
+esac
+
+shift
+
+for url; do
+    download_dir=$(mktemp -d)
+
+    echo Downloading $(basename "$url")...
+    wget -P $download_dir -o $download_dir/wget.log "$url" || (cat $download_dir/wget.log && false)
+    tar --no-same-owner -xf $download_dir/*.tar* -C $dest
```
ericonr, line 22: Couldn't we use `$download_dir/*` here? This doesn't support stuff like `.tgz` files.

henriquesimoes: We can if we log `wget` elsewhere or remove the log file afterwards. It might be a better to do so than to restrict the filename indeed. I'll update it. :)

ericonr (approved):

PR #22, opened 2023-10-09
ioc: use non-interactive install for build-stage.

henriquesimoes:
This is required to use `BUILD_PACKAGES` to install any package. No recommended package option has also been enabled to avoid unused packages in the resulting IOC image. Lines have been broken to make it easier to read and maintain this step.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,10 @@ ARG REPONAME
 ARG BUILD_PACKAGES
 ARG BUILD_TAR_PACKAGES
 
-RUN if [ -n "$BUILD_PACKAGES" ]; then apt update && apt install $BUILD_PACKAGES; fi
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -35,7 +35,11 @@ FROM build-image AS build-stage
 ARG REPONAME
 ARG BUILD_PACKAGES
 
-RUN if [ -n "$BUILD_PACKAGES" ]; then apt update && apt install $BUILD_PACKAGES; fi
+RUN if [ -n "$BUILD_PACKAGES" ]; then \
+    apt update && \
+    apt install -y --no-install-recommends \
+        $BUILD_PACKAGES; \
```
ericonr, line 41: Please keep these in the same line.

ericonr (changes requested):

PR #25, opened 2023-10-27
ioc: add script to launch IOCs inside container.

ericonr:
Setting this script as the default entrypoint for our containers allows IOCs which use "st.cmd" as their startup script to be launched without specifying any command in the command line or YAML orchestrator file, and still support other startup scripts or shell wrappers that perform the necessary setup before launching the IOC, but not having to call procServ themselves.

Furthermore, it means that there is a default location for the procServ socket, which can be documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,10 @@ you **need** to link them dynamically, you must define the build target as
 `dynamic-link`. This will increase the resulting image size, since unused
 dependencies will also be copied.
 
+The resulting image contains a standard IOC run script, `lnls-run`, which will
+be run inside `RUNDIR` and will launch the container's command under procServ,
+or `st.cmd` if no command is specified.
+
 Some Docker versions don't use
 [BuildKit](https://docs.docker.com/build/buildkit/) by default, and it can be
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -70,3 +70,16 @@ not built in `ADSupport`.
 
 Known build and runtime issues are documented in the [SwC
 wiki](http://swc.lnls.br/).
+
+## Containers
+
+### Accessing `iocsh` inside containers
+
+If a container is using the default `lnls-run` entrypoint (i.e. this won't work
```
henriquesimoes, line 78: Shouldn't we specify close to the `docker-compose.yml` example that we have a default entrypoint that runs procServ and expects an optional argument that is a iocsh script in `RUNDIR`?

I'm worried about people not noticing that they do not need to necessarily write a entrypoint script anymore.

henriquesimoes (commented):
Seems good to me. Just some documentation notes.

README.md
henriquesimoes: Seems good to me. Thanks!

henriquesimoes (approved):

PR #26, opened 2023-10-27
base: add support for Pmac module.

guirodrigueslima:


base/.env
```diff
@@ -10,5 +11,8 @@ AUTOSAVE_VERSION=R5-10-2
 SSCAN_VERSION=R2-11-5
 RECCASTER_VERSION=1.5
 IPAC_VERSION=2.16
+
 AREA_DETECTOR_VERSION=R3-12-1
+
 MOTOR_VERSION=R7-3-1
+PMAC_VERSION=2-6-0
```
henriquesimoes, line 18: Please, keep the newline at the end-of-file.

base/Dockerfile
```diff
@@ -17,7 +17,8 @@ RUN apt update -y && \
         libxml2-dev \
         re2c \
         wget \
-        ca-certificates
+        ca-certificates \
+        libssh2-1-dev
```
henriquesimoes, line 21: Let this between `libreadline-dev` and `libtiff-dev`. They are alphabetically sorted so that it is easier to identify if a package is already installed.

base/install_motor.sh
```diff
@@ -34,3 +33,45 @@ cd ..
...
+cd pmac
+
+rm -rf .git/ .git*
+
+module_releases="PMAC=${EPICS_MODULES_PATH}/pmac
+"
+echo "$module_releases" >> ${EPICS_MODULES_PATH}/../RELEASE
+
+echo "
+" > configure/RELEASE.local.linux-x86_64
```
henriquesimoes, line 53: I'd rather just `rm -f` it.

IMHO, we may remove any other `*linux*x86` file that influences the build. Special care with architecture-dependent configuration can be handled when we start testing `epics-in-docker` with each of them. For now, Intel x86-64 is our default.

base/install_motor.sh
```diff
@@ -34,3 +33,45 @@ cd ..
...
+
+echo "
+EPICS_BASE=/opt/epics/base
```
henriquesimoes, line 56: Please, use `${EPICS_BASE_PATH}`.

base/install_motor.sh
```diff
@@ -34,3 +33,45 @@ cd ..
...
+BUILD_IOCS=NO
+USE_GRAPHICSMAGICK=NO
+
+SSH             = $(pkg-config --variable=prefix libssh2)
+SSH_LIB         = $(pkg-config --variable=libdir libssh2)
+SSH_INCLUDE     = -I$(pkg-config --variable=includedir libssh2)
+
+WITH_BOOST = NO
+
+" > configure/CONFIG_SITE.linux-x86_64.Common
```
henriquesimoes, line 73: I'd rather remove `CONFIG_SITE.linux-x86_64.Common` and output to `CONFIG_SITE.local` instead. Same rationale from the comment above.

base/install_motor.sh
```diff
@@ -34,3 +33,45 @@ cd ..
...
+
+make -j${JOBS}
+make clean
+
```
henriquesimoes, line 77: Keep a single newline at the end-of-file.

base/install_motor.sh
```diff
@@ -34,3 +33,45 @@ cd ..
...
+#Add pmac module
+
+cd ..
+
+git clone --depth 1 --branch ${PMAC_VERSION} \
+    https://github.com/dls-controls/pmac
+
+cd pmac
+
+rm -rf .git/ .git*
```
henriquesimoes, line 46: `rm -rf .git*` will do the job for `.git` as well. `*` may expand to empty string and it will still match.

henriquesimoes (commented):
As we talked in person, `pmac` may be installed with `install_github_module` (or even with `download_github_module` and `install_module`). Refactor to simplify things when #27 is merged.

I've found some points to improve nevertheless. You may work on them when refactoring changes.

base/install_motor.sh
guirodrigueslima: Eu tentei remover, mas foi gerado erro na compilação, vou procurar o arquivo que inclui o `RELEASE.local.linux-x86_64` e comentar esse trecho.

base/install_motor.sh
guirodrigueslima: The `CONFIG_SITE` file does not include `CONFIG_SITE.local`. To do this we would have to remove `CONFIG_SITE.linux-x86_64.Common`, edit `CONFIG_SITE` and create `CONFIG_SITE.local`. Would we be making things more confusing?

henriquesimoes: I like the implementation you did writing directly to `CONFIG_SITE` instead.

base/install_motor.sh
```diff
@@ -34,3 +35,33 @@ cd ..
...
+rm -f pmac/configure/RELEASE.linux-x86_64.Common
+rm -f pmac/configure/CONFIG_SITE.linux-x86_64.Common
+
+echo "
+BUILD_IOCS=NO
+USE_GRAPHICSMAGICK=NO
+
+SSH             = $(pkg-config --variable=prefix libssh2)
+SSH_LIB         = $(pkg-config --variable=libdir libssh2)
+SSH_INCLUDE     = -I$(pkg-config --variable=includedir libssh2)
```
henriquesimoes, line 55: @ericonr, do you have any clue why `--cflags-only-I` wouldn't work even with variables defined in `libssh2.pc`?

I've compared the syntax along with @guirodrigueslima, but I couldn't understand what was happening. In this case, it is okay to have `includedir`, since it has a single include directory.

henriquesimoes (changes requested):
I don't think you need to specify the version in your commit message.

I'm missing an explanation in the message about the linux-x86_64 configuration files we removed. Explaining why we used `--variable=includedir` instead of `--cflags-only-I` may also be mentioned.

Indicating that the compilation error is probably due to not well-defined dependencies in the module's Makefile might be worth it to make this detail clearer.

henriquesimoes (commented):
Looks good for me. Waiting for @ericonr's opinion on that.

base/install_motor.sh
```diff
@@ -34,3 +36,33 @@ cd ..
 
 make -j${JOBS}
 make clean
+
+#Add pmac module
+
+cd ..
```
ericonr, line 42: `cd  $EPICS_MODULES_PATH`

base/install_motor.sh
```diff
@@ -34,3 +36,33 @@ cd ..
 
 make -j${JOBS}
 make clean
+
+#Add pmac module
```
ericonr, line 40: I don't know if this comment is needed, since we are explicitly talking about pmac right below :thinking:

base/install_motor.sh
```diff
@@ -34,3 +36,33 @@ cd ..
...
+
+cd ..
+
+download_github_module dls-controls pmac $PMAC_VERSION
+
+rm -f pmac/configure/RELEASE.local.linux-x86_64
+rm -f pmac/configure/RELEASE.linux-x86_64.Common
+rm -f pmac/configure/CONFIG_SITE.linux-x86_64.Common
```
ericonr, line 48: I'd prefer to not have the `-f` flag. This step will always run on a clean extraction of the module, and it would be good to get errors if any of these files are removed/renamed.

base/install_motor.sh
```diff
@@ -34,3 +36,33 @@ cd ..
...
+BUILD_IOCS=NO
+USE_GRAPHICSMAGICK=NO
+
+SSH             = $(pkg-config --variable=prefix libssh2)
+SSH_LIB         = $(pkg-config --variable=libdir libssh2)
+SSH_INCLUDE     = -I$(pkg-config --variable=includedir libssh2)
+
+WITH_BOOST = NO
+
+" >> pmac/configure/CONFIG_SITE
```
ericonr, line 60: Nit: don't need this empty line in the middle

(if it was because of the newline comment, `echo` will add it by default)

ericonr (changes requested):
I will try to take a look at the `pkg-config` issue. It's likely that the package is available in the default includedir, so there's no need to actually define any cflag for it.

I believe this PR is missing documentation update (in the README) mentioning the new `libssh2` dependency for IOCs using the Pmac module.

If possible, formatting the commit message body to fit in 72 columns would also be nice (can be done inside Vim in `git commit --amend` dialog by selecting the body and pressing `gq`). Otherwise I can do it myself before merging.

henriquesimoes (commented):
I was almost forgetting, but since `v0.4.0` we should add an entry in [CHANGES.md](https://github.com/cnpem/epics-in-docker/blob/v0.4.0/CHANGES.md#unreleased) under Unreleased section indicating this PR under "New features".

base/install_motor.sh
```diff
@@ -34,3 +36,30 @@ cd ..
...
+rm pmac/configure/RELEASE.linux-x86_64.Common
+rm pmac/configure/CONFIG_SITE.linux-x86_64.Common
+
+echo "
+BUILD_IOCS=NO
+USE_GRAPHICSMAGICK=NO
+
+SSH             = $(pkg-config --variable=prefix libssh2)
+SSH_LIB         = $(pkg-config --variable=libdir libssh2)
+SSH_INCLUDE     = -I$(pkg-config --variable=includedir libssh2)
```
ericonr, line 54: In this case I'd either leave it as `pkg-config --cflags-only-I libssh2` (the empty output might change on some other distro? who knows) or empty.

ericonr (changes requested):

base/install_motor.sh
```diff
@@ -34,3 +36,30 @@ cd ..
...
+
+rm pmac/configure/RELEASE.local.linux-x86_64
+rm pmac/configure/RELEASE.linux-x86_64.Common
+rm pmac/configure/CONFIG_SITE.linux-x86_64.Common
+
+echo "
+BUILD_IOCS=NO
+USE_GRAPHICSMAGICK=NO
+
+SSH             = $(pkg-config --variable=prefix libssh2)
```
ericonr, line 52: And consider making this `SSH=YES`

base/install_motor.sh
henriquesimoes: Agreed. Having it set by `pkg-config` seems better for me.

base/install_motor.sh
henriquesimoes: I've thought about that too. We should just remember that `SSH=NO` doesn't mean what it looks like =P.

henriquesimoes (approved):

ericonr (approved):

PR #27, opened 2023-10-27
base: extract install functions to shared file.

henriquesimoes:
These installation functions can be used in other scripts. This simplifies their logic through reuse and centralizes standardization. I've also refactored `install_epics.sh` to use it already.

The name of the script is something I'm open to suggestions. I'm okay with the name, but I'm sure there is a better one that I couldn't know of. :)

### DIFF
--- a/base/install_modules.sh
+++ b/base/install_modules.sh
@@ -2,43 +2,7 @@
 
 set -ex
 
-download_github_module() {
-    github_org=$1
-    module_name=$2
-    tag=$3
-
-    lnls-get-n-unpack -l https://github.com/$github_org/$module_name/archive/refs/tags/$tag.tar.gz
-
-    mv $module_name-$tag $module_name
-}
-
-install_module() {
-    module_name=$1
-    dependency_name=$2
-    release_content="$3"
-
-    cd $module_name
-    echo "$release_content" > configure/RELEASE
-
-    make -j${JOBS} install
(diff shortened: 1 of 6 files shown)


base/install_modules.sh
```diff
@@ -2,43 +2,7 @@
...
-    github_org=$1
-    module_name=$2
-    dependency_name=$3
-    tag=$4
-    release_content="$5"
-
-    download_github_module $github_org $module_name $tag
-    install_module $module_name $dependency_name "$release_content"
-}
+. ./install-functions.sh
```
ericonr: If epics-base is using an absolute path, why isn't this one?

Edit: I see you fix that in the following commit; please squash it into this first one.

ericonr (changes requested):
For the second commit, I believe it should be split into two or three:

- creating and using `EPICS_RELEASE_FILE`
- using `PWD` in `install_module`
- using `install_module` for base, and cleaning up the definition from `install_modules.sh`, which IMO a commit message should point out was the wrong place from the start

I think it would work for the last two to be the same commit as well, since PWD support was necessary to allow base to use the function.

base/install_modules.sh
henriquesimoes: Oops. Let this pass by mistake. Good catch.

ericonr (approved):

PR #28, opened 2023-10-30
base: apply StreamBuffer::grow patch to StreamDevice.

henriquesimoes:
This patch is applied with git-apply(1) so that the Git history is not required to exist. This fix should be removed once it is accepted in upstream.

A directory for patches has been created so that they aren't scattered across the installation scripts' directory.


henriquesimoes:
My mistake: this actually doesn't fix the bug which leads to the segmentation fault. I'll keep track of the upstream discussion to see if I close this PR or update it to the actual fix.

ericonr:
To make finding it easier: https://github.com/paulscherrerinstitute/StreamDevice/pull/96

ericonr:
As discussed in the PR, that isn't the right fix. So closing this PR until further investigation.

PR #29, opened 2023-11-08
base: add caPutLog module.

ericonr:


base/.env
```diff
@@ -12,3 +12,4 @@ RECCASTER_VERSION=1.5
 IPAC_VERSION=2.16
 AREA_DETECTOR_VERSION=R3-12-1
 MOTOR_VERSION=R7-3-1
+CAPUTLOG_VERSION=R4.0
```
henriquesimoes, line 15: This list is starting to get big. Maybe we should better define how it is organized. As of now, we followed the same order modules are built, right?

To keep this logic, this should go right after IPAC module. But I'm not sure if that order useful nevertheless.

henriquesimoes (approved):

Issue #16, opened 2023-09-04
Document changes


ericonr:
Changes include things such as:

- packages which need to be installed for kinds of IOC
- new/changed environment variables
- new/changed docker compose directives

henriquesimoes:
I've never maintained a changelog before. Do you think it might be interesting to follow a well-defined pattern, such as [Common Changelog](https://common-changelog.org/)?

ericonr:
I've never maintained a structured one either. I will take a look at that pattern, thanks!

PR #30, opened 2023-11-08
RELEASE: document release procedure.

ericonr:


RELEASE.md
```diff
@@ -0,0 +1,52 @@
...
+### Create release
+
+After pushing the resulting tag, a GitHub release [should be
+created](https://github.com/cnpem/epics-in-docker/releases/new). It should have
+the same title as the tag, and the description can use the "Generate release
+notes" feature, but special care should be taken to split the merged PRs into
+the "Breaking changes", "New features", and "Bug fixes" categories. The first
+section of the release description should be "Who needs this update" and
+describe the most relevant changes and package updates and what their effect
+is.
```
henriquesimoes, line 42: How about keeping the list from #16 explicit here? It might make it easier to look after specific (breaking) changes in large change sets.

henriquesimoes (commented):
Will #16 be closed by this PR? At first, I've thought about maintaining a `CHANGELOG.md` to track changes.

Will we keep that only in GitHub releases? That wouldn't go with the Git history if we move to another repository host.

RELEASE.md
```diff
@@ -0,0 +1,52 @@
...
+
+## `release` branch
+
+The `release` branch should be updated to point to the new tag.
+
+## `main` branch
+
+A new commit should be created in the `main` branch, updating the version used
+in [Dockerfile](./Dockerfile) to `vX.Y.Z-git` to avoid usage of the `main`
+branch in actual deployments.
```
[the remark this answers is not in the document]
ericonr: > Post-release development version bump

I'd be okay with something like that. Will add it.

I didn't remember what version we had settled on. I'm okay with `-dev`.

RELEASE.md
ericonr: That's a good idea, will add it.

ericonr:
> Will https://github.com/cnpem/epics-in-docker/issues/16 be closed by this PR? At first, I've thought about maintaining a CHANGELOG.md to track changes.

Is there tooling to automatically generate the changes?

I had started writing a changelog myself, but it's a lot of manual work. And having all commits misses out on the grouping by PRs, which I find useful. And having a link to the discussion about a given change, instead of only the commit, seems more useful to me as well. That's why I thought about relying on github releases.

henriquesimoes:
> Is there tooling to automatically generate the changes?

Not one that I'm familiar with. [github-changelog-generator](https://github.com/github-changelog-generator/github-changelog-generator) has [a list of alternatives](https://github.com/github-changelog-generator/github-changelog-generator/wiki/Alternatives) for doing that.

> I had started writing a changelog myself, but it's a lot of manual work. And having all commits misses out on the grouping by PRs, which I find useful. And having a link to the discussion about a given change, instead of only the commit, seems more useful to me as well. That's why I thought about relying on github releases.

I agree. As we discussed in-person, let's stick with progressive writing of the changelog grouped by PR.

PR #31, opened 2023-11-08
base: add NDSSCPimega areaDetector plugin. 

henriquesimoes:
This plugin computes the geometric restoration for Pimega detectors based on the LNLS [ssc-pimega library](https://github.com/cnpem/ssc-pimega), and allows Pimega IOC to provide coherent data to users.

Even though ADPimega is not yet supported (#15), this plugin can still be used with ADSimDetector for debugging purposes. Furthermore, it makes it possible to build ADPimega (and its IOC) out of epics-in-docker, but still using NDSSCPimega from base.

This is now possible to be merged due to the refactoring I've done to use ssc-pimega from GitHub's repository. To reuse the built library in the runtime image, I've included the copy of `/usr/local/lib`.


base/.env
```diff
@@ -11,4 +11,7 @@ SSCAN_VERSION=R2-11-5
 RECCASTER_VERSION=1.5
 IPAC_VERSION=2.16
 AREA_DETECTOR_VERSION=R3-12-1
+NDSSCPIMEGA_VERSION=1.0.0
 MOTOR_VERSION=R7-3-1
+
+LIBSSCPIMEGA_VERSION=main
```
henriquesimoes, line 17: This is an unfortunate configuration. I'll see if we can have a tag available for downloading with `download_github_module`.

Dockerfile
```diff
@@ -33,6 +33,9 @@ ENTRYPOINT ["./entrypoint"]
 
 FROM base AS no-build
 
+COPY --from=build-image /usr/local/lib /usr/local/lib
+RUN ldconfig
+
```
henriquesimoes, line 38: I'm not sure if this is the best place to put it. I'm thinking about we applying an optimization afterwards, which will be applicable only for `no-build` and `dynamic-link`. But as of now, it is technically possible be moved to `base` (and also be applied to `static-link`), which avoids duplicated configuration.

Maybe it is time to create another shared stage for `no-build` and `dynamic-link`(?) This has some drawbacks, such as being required to duplicate packages in `{BUILD,RUNTIME}_TAR_PACKAGES` once again (due to using `COPY --from=build-image` in this shared stage).

base/.env
henriquesimoes: Switching this PR to draft until I fix this for non-rolling module versioning.

Dockerfile
ericonr: Do you mean something like `dynamic-base`? I don't know if it's worth it.

base/install_area_detector.sh
```diff
@@ -96,3 +98,17 @@ git apply --directory ADSupport ${EPICS_MODULES_PATH}/nanohttp_stream.patch
 
 make -j${JOBS}
 make clean
+
+cd ..
+
+git clone https://github.com/cnpem/ssc-pimega
+git -C ssc-pimega checkout $LIBSSCPIMEGA_VERSION
+make -C ssc-pimega/c install
```
ericonr, line 106: It is possible to download the tarball for a specific git hash, and IMO we should do that.

ericonr (changes requested):

base/install_area_detector.sh
henriquesimoes: I like the idea. Would you rather support hash in `download_github_module` or use a directly link for this case in particular?

Supporting only tags seems better to force us to use only released code. On the other hand, it makes it harder to handle such cases in which the project does not use tags as a way to release.

Dockerfile
henriquesimoes: It could be. I don't know either if its worth either.

A relevant point I now see is that it might be interesting for `static-link` as well. Any system library we build from source will require to be available in the IOC image, even with `STATIC_BUILD=YES`.

base/install_area_detector.sh
ericonr: > Supporting only tags seems better to force us to use only released code. On the other hand, it makes it harder to handle such cases in which the project does not use tags as a way to release.

That's a debate I have seen a lot of times, lol

Extending the function to support hashes would be best if we know we are going to depend on it. Who's to say we are never going to need bug fixes for which a patch isn't feasible? I think I'd go for it.

Dockerfile
henriquesimoes: As we discussed in person, I moved it to `base`.

ericonr (approved):

CHANGES.md
```diff
@@ -2,10 +2,16 @@
 
 ## Unreleased
 
+<<<<<<< HEAD
```
henriquesimoes, line 5: Oops

PR #33, opened 2023-11-17
Separate pimega build from area_detector.

marcomontevechi1:
Pimega build is something that we will probably do frequently here and throwing away the areadetector build cache to rebuild pimega is time consuming.

Maybe this could be done for all modules? Particularly for me, pimega is enough, but standardization is never bad.


marcomontevechi1:
Closing the request as discussed in person.
Local developments can be made while the definitive version still exists.
In the long run it still makes sense to leave support areaDetector modules in epics-in-docker

PR #36, opened 2023-11-30
base: build epics-base with GCC_PIPE=YES.

ericonr:
This can speed up the build by avoiding usage of temporary files.


ericonr:
I'd say this doesn't need an entry in changes :thinking: 
What do you think?

ericonr:
From looking around online, it doesn't help a whole lot, and can make memory consumption worse. Will close this

PR #38, opened 2023-11-30
ci: use docker/bake-action.

ericonr:
This allows us to avoid writing our own commands for things, and works with GHA cache for images, which speeds up CI and avoids unnecessary resource usage.

### DIFF
--- a/.github/workflows/base-image.yml
+++ b/.github/workflows/base-image.yml
@@ -25,17 +25,21 @@ jobs:
       IMAGE: ${{ matrix.image.name }}
       SOURCE: https://github.com/${{ github.repository }}
     steps:
-      - uses: actions/checkout@v3
-      - name: Build image
-        run: cd ${{ matrix.image.dir }} && JOBS=$(nproc) docker compose build
+      - uses: actions/checkout@v4
+      - uses: docker/setup-buildx-action@v3
       - name: Log in to registry
         if: github.event_name == 'push' && vars.PUSH_TO_REGISTRY == 'true'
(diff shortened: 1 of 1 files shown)


ericonr:
As is, this PR means we wouldn't publish latest images any more. I am okay with it, but I don't know if there are use cases I'm not thinking of.

henriquesimoes:
> [run without cache](https://github.com/cnpem/epics-in-docker/actions/runs/7105624218). which should have populated cache for the [next run](https://github.com/cnpem/epics-in-docker/actions/runs/7106041544)

Is there any reason why musl-based image build didn't use caching in second linked run? Have this already been fixed or this is an occasional issue?

> As is, this PR means we wouldn't publish latest images any more. I am okay with it, but I don't know if there are use cases I'm not thinking of.

I can't think of a user scenario where this would be required. Actually, it might be good not to provide it, aiming to prevent locally changed Dockerfile still build against an (ephemeral) registry image.

ericonr:
> Is there any reason why musl-based image build didn't use caching in second linked run? Have this already been fixed or this is an occasional issue?

That was fixed with the latest CI runs. The issue was that the scope wasn't specified, and the cache is fully uploaded every CI run, so if a scope isn't specified, CI runs overwrite each other.

henriquesimoes:
> Should we consider deleting the latest images?

I'd go for it. Only the latest `latest` image still exists, right? Is there any previous one still available somehow?

ericonr:
The other images are still in the registry, but they can only be referenced by their hash.

henriquesimoes:
> [...] they can only be referenced by their hash.

Oh, right. I didn't see the "Untagged" tab.

henriquesimoes (approved):

PR #39, opened 2023-12-04
Release version v0.5.0.

henriquesimoes:
The corresponding tag should be included after merging this PR.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+## v0.5.0
+
+Users willing to use Pmac or NDSSCPimega should update to this release. A
+musl-based image has been added, enabling support for linking fully static
+binaries for deployment outside of containers.
+
 ### New features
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+## v0.5.0
+
+Users willing to use Pmac or NDSSCPimega should update to this release. Users
+interested in fully static linked binaries through musl build should update as
+well.
+
```
ericonr, line 10: I don't think we should mention the musl build this way, because it was a use case we didn't cover, so updating to it doesn't make sense, IMO.

"A musl-based image has been added, enabling support for linking fully static binaries for deployment outside of containers."

henriquesimoes: Nice. That way it is clearer indeed what this is for. Thanks!

ericonr (approved):

PR #43, opened 2023-12-18
Update 2023-12-08

ericonr:


henriquesimoes:
Which approach we will use regarding the `CHANGES.md`? Breaking changes that are likely to affect our IOCs should be listed, IMO. But how about the other changes? We are going to mention it as an "update \<module> to \<version>" or not even that?

henriquesimoes:
>  I tried to understand what the resource leak fix in autosave was, but couldn't find the proper commit

There seems to be several commits, actually. I've found some of them, which were found by SonarQube static analysis.

https://github.com/epics-modules/autosave/compare/666bf2f495212d91d34eb26f8eea86bdbf231e54...90abf19cc3968217003cba30fa74659452e56f1c

Some commits ahead seems to fix other reports from the same tool, but does not explicitly mention resource leakage in the commit message.

> if it is an actual resource leak, I'd like to add a recommendation for all users to update, even.

That would be nice.

ericonr:
> There seems to be several commits, actually. I've found some of them, whose leaks were found by SonarQube static analysis.

Oh, nice. Thanks! I was looking at the commit list on github, didn't look at PRs or merge commits. And hadn't noticed the time since the last release either.

About sscan, I managed to find a tool for searches on GitLab, and didn't find anyone using `scan_settings.req` in the SOL group. However, I found some modules on GitHub which use it (I don't think they matter for us though):

- https://github.com/epics-modules/dxp
- https://github.com/kmpeters/electronAnalyserViewer
- https://github.com/CARS-UChicago/GSECARS-EPICS
- https://github.com/tomography/tomoscan
- https://github.com/Engbretson/laserTracker
- https://github.com/epics-modules/std

AreaDetector autosave file mentions it, but it's commented out.

henriquesimoes:
> About sscan, I managed to find a tool for searches on GitLab, and didn't find anyone using scan_settings.req in the SOL group. However, I found some modules on GitHub which use it (I don't think they matter for us though):

Good. It should be straightforward to update it, then.

> AreaDetector autosave file mentions it, but it's commented out.

Well, that is also an example script. I'd expect areaDetector users to write their own version instead of including that. But if anyone had done so, it would appear in your search.

ericonr:
@henriquesimoes could you review this again so we can merge it?

henriquesimoes (approved):
It's fine by me.

PR #45, opened 2023-12-19
base: disable bash tracing from install scripts.

henriquesimoes:
This level of detail pollutes the output during the build, while often not being used. In addition, it can be easily turned back on in a local setup for debugging purposes.

It should not influence any output seen by epics-in-docker users, so no entry is specified in the changelog.

### DIFF
--- a/base/install_area_detector.sh
+++ b/base/install_area_detector.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -ex
+set -e
 
 . /opt/epics/install-functions.sh
 
--- a/base/install_epics.sh
+++ b/base/install_epics.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-set -ex
+set -e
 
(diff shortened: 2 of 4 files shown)


ericonr:
I don't know if I agree with this. Most of the time I'm not looking at the build output anyway, and when there are issues, I don't need to run anything again, the trace output is there already. Has it bothered you in some situation?

henriquesimoes:
That's something I've added during the development of the installation scripts back when it was a sketch for containerizing the RFFE IOC.

I think I misunderstood from one of our personal discussion that it wouldn't be needed, similarly to logic of omitting the verbose output from `tar`.

> Has it bothered you in some situation?

Not really. We may keep them. ;)

ericonr:
> I think I misunderstood from one of our personal discussion that it wouldn't be needed, similarly to logic of omitting the verbose output from tar.

I might have changed my mind on it in the meanwhile, to be quite honest :thinking:

PR #48, opened 2024-01-08
base: use RELEASE to determine module paths.

henriquesimoes:
Right after being installed, a module adds its own path to the RELEASE file. Use this as the source of truth for specifying the dependencies' path in later module builds. Since a module cannot be built without its dependency already built, this does not limit our installation scripts.

This removes duplication of several paths of modules, which may now change more easily.

### DIFF
--- a/base/install-functions.sh
+++ b/base/install-functions.sh
@@ -1,3 +1,11 @@
+get_module_path() {
+    for module in $@; do
+        if [ -n "$module" ]; then
+            grep -E "^$module=" $EPICS_RELEASE_FILE
+        fi
+    done
+}
+
 download_github_module() {
     github_org=$1
     module_name=$2
@@ -11,10 +19,11 @@ download_github_module() {
 install_module() {
     module_name=$1
     dependency_name=$2
-    release_content="$3"
+    release_modules="$3"
 
     cd $module_name
-    echo "$release_content" > configure/RELEASE
+    get_module_path "$release_modules" > configure/RELEASE
+
     if [ -n "$NEEDS_TIRPC" ]; then
         echo "TIRPC=YES" >> configure/CONFIG_SITE
     fi
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,8 @@
 * base: update modules. by @ericonr in
   https://github.com/cnpem/epics-in-docker/pull/43
   * This updates the autosave module and removes some resource leaks.
+* base: build CALC with Sequencer support. by @henriquesimoes in
+  https://github.com/cnpem/epics-in-docker/pull/48
 
 ## v0.6.0
 
--- a/base/install_area_detector.sh
+++ b/base/install_area_detector.sh
@@ -29,15 +29,13 @@ ADCORE=${EPICS_MODULES_PATH}/areaDetector/ADCore
 
 echo "$module_releases" >> ${EPICS_RELEASE_FILE}
 
-echo "
-EPICS_BASE=${EPICS_BASE_PATH}
-
-$module_releases
-
-ASYN=${EPICS_MODULES_PATH}/asyn
-BUSY=${EPICS_MODULES_PATH}/busy
-SSCAN=${EPICS_MODULES_PATH}/sscan
+get_module_path "
+EPICS_BASE
+ASYN
+BUSY
+SSCAN
 " > RELEASE.local
+echo "$module_releases" >> RELEASE.local
 
 ln -s RELEASE.local RELEASE_PRODS.local
 ln -s RELEASE.local RELEASE_LIBS.local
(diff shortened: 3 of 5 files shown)


base/install-functions.sh
```diff
@@ -1,3 +1,11 @@
+get_module_path() {
+    for module; do
+        if [ -n "$module" ]; then
+            grep -E "^$module=" $EPICS_RELEASE_FILE
+        fi
```
henriquesimoes, line 5: I'm not sure if that's the best way to do it. I don't want to restrict the parameter to have only lines with content. This would require the modules to be all in the same line.

Let me know if you know way better way to handle this type of parameter.

ericonr: This works and doesn't seem overly inefficient, so I don't see any issue with it.

What do you think about supporting manually specified dependencies anyway (check if there's a `=`, if yes, just pass through the whole line)? Some weird case might require it (but then it could also be more correct to fix epics-in-docker to include that specific dep in the release file).

henriquesimoes: > This works and doesn't seem overly inefficient, so I don't see any issue with it.

Great! The previous version of it was worse and buggy, nice to have it right at last.

> What do you think about supporting manually specified dependencies anyway (check if there's a =, if yes, just pass through the whole line)?

I had thought about that, but in a different way. I would change the install function to conditionally overwrite configure/RELEASE before installing based on the emptiness of the module list. But I dropped the idea when I thought to myself whether we wanted or not to enforce all installed modules to be listed there.

> Some weird case might require it

Maybe a module that requires (a submodule of) itself? Not sure how wild module dependencies can become.

ericonr: > Maybe a module that requires (a submodule of) itself? Not sure how wild module dependencies can become.

I dearly hope that's not the case... I think we can omit it for now.

ericonr (commented):
Nits:

"base: use RELEASE to determine module paths."

"..., each module adds its..."
"Therefore, it's possible to use the RELEASE file as the source of truth for our dependencies' paths"

"This removes duplication of several module paths, which may now be changed more easily."

That last phrase makes me feel like we are going to be changing a whole lot of paths :p Is that why you mentioned this being related to the final container cleanup?

henriquesimoes:
> Nits:

Thanks!

> That last phrase makes me feel like we are going to be changing a whole lot of paths :p Is that why you mentioned this being related to the final container cleanup?

Yep. That's the point. Not sure if we'll really need to move things around, but it is now easier to do so if the decide this is the best option. Nevertheless, having a single point for defining such things diminishes our chances of making mistakes. ;)

base/install-functions.sh
```diff
@@ -1,3 +1,11 @@
...
+    done
```
ericonr, line 6: Ok, I'm now a bit confused... This shouldn't be working, and yet the build succeeds.

You are passing the arguments as a quoted list; that means your `for` loop runs only once, and searches for all the modules in a single grep, and shouldn't find any matches... Example from CI run:

```
2024-01-24T20:38:45.4438379Z #19 249.2 + grep -E '^
2024-01-24T20:38:45.4438645Z #19 249.2 EPICS_BASE
2024-01-24T20:38:45.4438981Z #19 249.2 ASYN
2024-01-24T20:38:45.4439209Z #19 249.2 CALC
2024-01-24T20:38:45.4439485Z #19 249.2 =' /opt/epics/RELEASE
```

For it to work as you expected, do `for module in $@;`.

Since we are making changes here, I think we should also error out if the output is empty, as will be the case for the wrong module name for sequencer (SEQ) when building CALC (this PR should then also fix that in a first commit).

henriquesimoes: > Ok, I'm now a bit confused... This shouldn't be working, and yet the build succeeds.

Yeah, it should break with no dependencies defined.

> Example from CI run

Nice to see that `-x` turned out to be really useful in the end. ;)

> For it to work as you expected, do for module in $@;.

Makes sense. I'll update it.

> I think we should also error out if the output is empty

That would mean we couldn't use it for EPICS base, right?

> this PR should then also fix that in a first commit

Good point. I did notice this in other context, but I forgot about that when making these changes.

henriquesimoes: > That would mean we couldn't use it for EPICS base, right?

I've checked this and it is not the case. We'll be passing an empty argument, which won't even trigger the `grep` search, successfully passing by.

> I think we should also error out if the output is empty

That's currently the case. Since we set `-e` shell option, `grep` can't return non-zero, which is what happens when it doesn't find a match.

ericonr: > That's currently the case. Since we set -e shell option, grep can't return non-zero, which is what happens when it doesn't find a match.

Please document it in the commit message :)

ericonr (approved):

PR #51, opened 2024-01-17
ci: add reusable job for IOC images.

ericonr:
Based on testing done for https://github.com/lnls-dig/rffe-epics-ioc/pull/11

### DIFF
--- a/.github/workflows/ioc-images.yml
+++ b/.github/workflows/ioc-images.yml
@@ -1,0 +1,35 @@
+name: Build image
+on:
+  workflow_call:
+
+jobs:
+  build_and_push:
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      contents: read
+    env:
+      REGISTRY: ghcr.io/${{ github.repository_owner }}
+      SOURCE: https://github.com/${{ github.repository }}
+      TAG: ${{ github.ref_name }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - uses: docker/setup-buildx-action@v3
+      - name: Log in to registry
+        if: github.event_name == 'push'
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push tagged image
+        uses: docker/bake-action@v4
+        with:
+          workdir: ./
+          files: docker-compose.yml
+          set: |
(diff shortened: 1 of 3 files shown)


.github/workflows/ioc-images.yml
```diff
@@ -0,0 +1,38 @@
...
+jobs:
+  build_and_push:
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      contents: read
+    env:
+      REGISTRY: ghcr.io/${{ github.repository_owner }}
+      SOURCE: https://github.com/${{ github.repository }}
+      TAG: ${{ github.ref_name }}
```
henriquesimoes, line 17: We should document that this must be specified in the `docker-compose.yml` image name so that the tagged image is pushed. Otherwise, only the `latest` image tag will be used.

Also change the template in the README.md to use it.

README.md
```diff
@@ -80,6 +85,38 @@ module depends on it.
...
+      - main
+  pull_request:
+
+jobs:
+  build_and_push:
+    permissions:
+      packages: write
+      contents: read
+    uses: cnpem/epics-in-docker/.github/workflows/ioc-images.yml
+    secrets:
```
ericonr, line 116: I'm not specifying `ioc-images.yml@v0...` here. It seems like a bad idea to add yet another place where users need to update version identifiers on update.

henriquesimoes: Sounds reasonable. That means will need to have the compromise to make it backward compatible, right?

ericonr: IMO breaking CI is not the end of the world. If we break anyone, they can then just pin the version if properly updating will be such a pain.

henriquesimoes: Agreed; looks good.

CHANGES.md
```diff
@@ -7,6 +7,9 @@
 * base: add PVXS module. by @ericonr in
   https://github.com/cnpem/epics-in-docker/pull/42
 
+* ci: add reusable job for IOC images. by @ericonr in
+  https://github.com/cnpem/epics-in-docker/pull/49
+
```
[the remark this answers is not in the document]
ericonr: AFAIK that's correct for markdown. My bad :)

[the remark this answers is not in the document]
henriquesimoes: Have you pushed changes for this?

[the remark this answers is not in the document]
ericonr: Now I have :innocent:

henriquesimoes (approved):

Issue #19, opened 2023-09-11
Prune unused modules from dynamic-link


henriquesimoes:
`dynamic-link` and `no-build` targets use a trivial copy all from `/opt`, leading to unused modules and unused build artifacts to be copied to the resulting IOC image.

Any level of pruning could easily and greatly improve the final image size.

_Discussion started by @ericonr in https://github.com/cnpem/epics-in-docker/pull/15#discussion_r1314850489_

ericonr:
Possible strategies:

- list of used modules in a docker-compose variable
- list of used modules taken from original RELEASE file before it's replaced
- prune directories at the `/opt/modules` and `/opt/modules/areaDetector` levels when they don't appear in `ldd`output for all executables/libraries in IOC / final directory (for `no-build` mode)

PR #55, opened 2024-03-25
ioc: copy locally installed software from build stages.

henriquesimoes:
Software, other than shared libraries, may be built in the first stages, but are currently not available in the resulting IOC image. Copy the entire `/usr/local` tree in all IOC stages, so that locally installed binaries and configuration can be usable at runtime.

This allows us to build custom versions of applications during build phases and install them at IOC images, centralizing and making it easier to track how they are built. In addition, IOCs can also install custom files which will be available at runtime in a known location. Still, IOCs themselves can also be installed in [FHS-compliant paths][1] inside the container as any other packaged software, allowing them to make use of tools based on system-installed artifacts (such as `env` executable discovery).

[1]: https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s09.html

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,10 @@
   https://github.com/cnpem/epics-in-docker/pull/42
 * ci: add reusable job for IOC images. by @ericonr in
   https://github.com/cnpem/epics-in-docker/pull/49
+* ioc: copy locally installed software from build stages. by @henriquesimoes in
+  https://github.com/cnpem/epics-in-docker/pull/55
+  * This allows the IOC build phase to install custom files, available at
+    runtime, under `/usr/local` tree for all targets.
 
 ### Bug fixes
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,7 @@ ENTRYPOINT ["./entrypoint"]
 FROM base AS no-build
 
 COPY --from=build-image /opt /opt
+COPY --from=build-image /usr/local /usr/local
 
 
 FROM build-image AS build-stage
@@ -70,6 +71,7 @@ RUN make distclean && make -j ${JOBS} && make clean && make -C ${RUNDIR}
 FROM base AS dynamic-link
 
 COPY --from=dynamic-build /opt /opt
+COPY --from=dynamic-build /usr/local /usr/local
 
 
 FROM build-stage AS static-build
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -8,6 +8,10 @@
   https://github.com/cnpem/epics-in-docker/pull/42
 * ci: add reusable job for IOC images. by @ericonr in
   https://github.com/cnpem/epics-in-docker/pull/49
+* ioc: copy locally installed software from build stages. by @henriquesimoes in
+  https://github.com/cnpem/epics-in-docker/pull/55
+  * This allows the IOC build phase to install custom files, available at
+    runtime, under `/usr/local` tree for all targets.
```
ericonr, line 14: I don't think "locally installed" is clear enough, because distro packages are still locally installed, and for those we should be using the package installation mechanism.

The same goes for the commit message.

Dockerfile
```diff
@@ -37,6 +37,7 @@ ENTRYPOINT ["./entrypoint"]
 FROM base AS no-build
 
 COPY --from=build-image /opt /opt
+COPY --from=build-image /usr/local /usr/local
```
ericonr, line 40: Don't we have scripts here which are only relevant for the image build phase? :thinking: 
Like `lnls-get-n-unpack`

ericonr (changes requested):

CHANGES.md
henriquesimoes: > I don't think "locally installed" is clear enough

I understand. I followed the terminology used by FHS itself:

> Locally installed software must be placed within /usr/local rather than /usr unless it is being installed to replace or upgrade software in /usr. [(In 4.9 Local hierarchy)](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s09.html#ftn.idm236091898144)

Any suggestion?

Dockerfile
henriquesimoes: You are right. But after-build stages wouldn't be aware of such things. We must remove everything (including `lnls-get-n-unpack`) beforehand if we don't want it to be in the final image, otherwise we won't shrink the image size (which is related to #19).

ericonr: So it makes sense to wait for #59 before merging this. What do you think?

CHANGES.md
ericonr: so that's done already, great.

Dockerfile
henriquesimoes: I'm okay with it.

ericonr:
> If we make this something handled by epics-in-docker itself, it wouldn't need it. (Probably such internal implementation would, though.)

Are you suggesting the IOC build script copy executables itself?

PR #57, opened 2024-04-10
Opcua IOC

guirodrigueslima:
- Added opca ioc to epics-in-docker
- The IOC was not build with the install_module() function, as "make clean" removes the envPaths file.


guirodrigueslima:
Added ether_ip IOC to PLCs Rockwell family.

base/install_opcua.sh
```diff
@@ -0,0 +1,29 @@
...
+
+echo "
+EPICS_BASE=${EPICS_BASE_PATH}
+" > configure/RELEASE.local
+
+make -j${JOBS}
+
+EPICS_HOST_ARCH=$(${EPICS_BASE_PATH}/startup/EpicsHostArch)
+ln -s ${EPICS_MODULES_PATH}/opcua/opcuaIocApp/libopcua.so ${EPICS_MODULES_PATH}/opcua/lib/${EPICS_HOST_ARCH}
+ln -s ${EPICS_MODULES_PATH}/opcua/opcuaIocApp/libopcua.so.0.9 ${EPICS_MODULES_PATH}/opcua/lib/${EPICS_HOST_ARCH}
```
henriquesimoes, line 29: Do we need to keep both links? `libopcua.so.0.9` is the actual binary and `libopcua.so` is already a symbolic link. I don't know if we are ever going to need to support multiple versions to coexist.

Still, I think this can be simplified if we use [bash brace expansion](https://www.gnu.org/software/bash/manual/html_node/Brace-Expansion.html):

```bash
ln -s ${EPICS_MODULES_PATH}/opcua/{opcuaIocApp/libopcua.so.0.9,lib/$EPICS_HOST_ARCH/libopcua.so}
```

base/install_opcua.sh
```diff
@@ -0,0 +1,29 @@
...
+
+cd opcua
+
+echo "
+EPICS_BASE=${EPICS_BASE_PATH}
+" > configure/RELEASE.local
+
+make -j${JOBS}
+
+EPICS_HOST_ARCH=$(${EPICS_BASE_PATH}/startup/EpicsHostArch)
```
henriquesimoes, line 27: `startup/EpicsHostArch` will be removed from [future releases of base][1]. You should call the Perl script directly.

```bash
perl ${EPICS_BASE_PATH}/lib/perl/EpicsHostArch
```

[1]: https://github.com/epics-base/epics-base/blame/cb1571783b4793cb9c545e5736659ee5ed7d03bb/startup/EpicsHostArch#L10

henriquesimoes (changes requested):

base/install-functions.sh
```diff
@@ -36,14 +36,36 @@ install_module() {
...
+
+    cd -
+}
+
 # Install module from GitHub tagged versions or URL
-install_github_module() {
+install_github() {
+    if [ "$1" == "-i" ]
+        then iocs=1; shift
+        else iocs=0
```
henriquesimoes, line 56: We can use `true` and `false` here instead. BTW, this should make it easier to write the `if` statement later:

```shell
is_ioc = true

if $is_ioc; then
  # do something...
fi
```

base/install_iocs.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -ex
+
+. /opt/epics/install-functions.sh
+
+#Build OPCUA IOC
+lnls-get-n-unpack -l https://github.com/epics-modules/opcua/releases/download/v${OPCUA_VERSION}/IOC_opcua-${OPCUA_VERSION}_Base-${EPICS_BASE_VERSION}_debian${DEBIAN_VERSION%.*}.tar.gz
```
henriquesimoes, line 8: Maybe make this line shorter somehow?

henriquesimoes (changes requested):
Besides the implementation, I think we can improve the messages by making them more descriptive and explanatory. For instance, why we are using another path for IOCs? Why do we now use another function to install things?

In addition, please let the first letter after the `:` lowercase and ending with a period (`.`), so that the message headline is kept in the pattern.

base/install_iocs.sh
guirodrigueslima: do you have some suggestions?

henriquesimoes: I was thinking in something like splitting it into a base URL definition and a filename, which are then used by `lnls-get-n-unpack`.

```bash

opcua_release_url=https://github.com/epics-modules/opcua/releases/download/v${OPCUA_VERSION}
opcua_release_file=IOC_opcua-${OPCUA_VERSION}_Base-${EPICS_BASE_VERSION}_debian${DEBIAN_VERSION%.*}.tar.gz

lnls-get-n-unpack -l $opcua_release_url/$opcua_release_file
```

It seems less cumbersome to read this way and easier to edit the path if needed, IMHO.

henriquesimoes:
> Besides the implementation, I think we can improve the messages by making them more descriptive and explanatory.

Sorry. I think I wasn't clear enough here. I'm talking about *commit* messages specifically. (Same thing for the the lowercase and period thing I said.)

base/install_iocs.sh
```diff
@@ -0,0 +1,21 @@
...
+EPICS_BASE
+"
+
+EPICS_HOST_ARCH=`perl ${EPICS_BASE_PATH}/lib/perl/EpicsHostArch.pl`
+ln -s ${EPICS_IOCS_PATH}/opcua/{opcuaIocApp/libopcua.so.0.9,lib/${EPICS_HOST_ARCH}/libopcua.so}
+
+#Build ether_ip IOC
+install_github -i epics-modules ether_ip ETHER_IP $ETHER_IP_VERSION "
+EPICS_BASE
+"
```
ericonr, line 21: This comment isn't necessary, and `ether_ip` should be installed in `install_modules.sh`, and its version should be grouped along with the modules installed on that script.

`ether_ip` addition should therefore be a separate commit.

CHANGES.md
```diff
@@ -8,6 +8,13 @@
   https://github.com/cnpem/epics-in-docker/pull/42
 * ci: add reusable job for IOC images. by @ericonr in
   https://github.com/cnpem/epics-in-docker/pull/49
+* base: new functionality for installing iocs, this feature was 
+  created to add default iocs to the epics-in-docker image and
+  be exported in the future as no-build, with this feature it was 
+  necessary to change some function name structures in the 
+  install-functions.sh file. 
+  added OPCUA and ether_ip IOC. by @guirodrigueslima in
+  https://github.com/cnpem/epics-in-docker/pull/57
```
ericonr, line 17: Suggestion: "base: add OPCUA and ether_ip modules. by ..."

base/install_iocs.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
ericonr, line 1: I'd name this file `install_opcua.sh` instead.

Even if we have new modules which install IOCs, not every platform will support opcua specifically, and it's easier to have it as a single script which we can omit from the Dockerfile.

base/install-functions.sh
```diff
@@ -36,14 +36,36 @@ install_module() {
...
+    cd -
+}
+
 # Install module from GitHub tagged versions or URL
-install_github_module() {
+install_github() {
+    if [ "$1" == "-i" ]
+        then is_ioc=true; shift
+        else is_ioc=false
+    fi
```
ericonr, line 57: Please keep the conditional structure we are already using (also use `=` for string comparisons):

```suggestion
    is_ioc=false
    if [ "$1" = "-i" ]; then
        is_ioc=true
        shift
    fi
```

base/install-functions.sh
```diff
@@ -36,14 +36,36 @@ install_module() {
...
     tag=$4
     release_content="$5"
 
-    download_github_module $github_org $module_name $tag
-    install_module $module_name $dependency_name "$release_content"
+    download_from_github $github_org $module_name $tag
+    if $is_ioc
+        then install_ioc $module_name $dependency_name "$release_content"
+        else install_module $module_name $dependency_name "$release_content"
+    fi
```
ericonr, line 69:
```suggestion
    if $is_ioc; then
        install_ioc $module_name $dependency_name "$release_content"
    else
        install_module $module_name $dependency_name "$release_content"
    fi
```

ericonr (changes requested):

base/install-functions.sh
```diff
@@ -36,14 +36,36 @@ install_module() {
...
+    get_module_path "$release_modules" > configure/RELEASE
+
+    make -j${JOBS} install
+
+    cd -
+}
+
 # Install module from GitHub tagged versions or URL
-install_github_module() {
+install_github() {
```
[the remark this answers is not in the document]
ericonr: This hasn't been done yet.

base/install_opcua.sh
```diff
@@ -0,0 +1,29 @@
...
+set -ex
+
+. /opt/epics/install-functions.sh
+
+lnls-get-n-unpack -l https://github.com/epics-modules/opcua/releases/download/v${OPCUA_VERSION}/IOC_opcua-${OPCUA_VERSION}_Base-${EPICS_BASE_VERSION}_debian${DEBIAN_VERSION%.*}.tar.gz
+mv binaryOpcuaIoc opcua
+
+#install_module opcua OPCUA "
+#EPICS_BASE=${EPICS_BASE_PATH}
+#"
```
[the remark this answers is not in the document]
ericonr: AFAIU we aren't cleaning the directory yet, are we?

base/install-functions.sh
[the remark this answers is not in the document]
guirodrigueslima: I changed download_from_github(), I will do it for install_from_github() too.

base/install_iocs.sh
guirodrigueslima: When I talked to @henriquesimoes, I thought that the iocs folder would be deleted when I built other iocs. Keeping the iocs.sh script isolated for epics-in-doker native iocs, could facilitate other applications in the future?

base/install_iocs.sh
henriquesimoes: > `ether_ip` should be installed in `install_modules.sh`

I like the idea of having it in a separate file if we have multiple IOCs being built in the base image, which distinguishes better support modules and other applications. For instance, musl-based image does not need to install `ether_ip`, because it won't generate a static library to be linked against another application.

Beyond that conceptual division, it is practical to have them installed somewhere else. Pruning can then [simply throw the entire root directory away](https://github.com/cnpem/epics-in-docker/pull/59#issuecomment-2059673694). This could still be scripted inside `install_module.sh` nevertheless.

guirodrigueslima:
Insert OPCUA ioc in alpine version ?

base/musl/Dockerfile
```diff
@@ -51,6 +51,8 @@ ARG RECCASTER_VERSION
 ARG IPAC_VERSION
 ARG CAPUTLOG_VERSION
 ARG RETOOLS_VERSION
+ARG OPCUA_VERSION
```
henriquesimoes, line 54: We don't have the static library for OPC-UA as far as I know. That means musl image won't be able to use build it anyway. Thus, we may remove this definition here and from its `docker-compose.yml`.

@ericonr, please correct me if I'm mistaken about this.

henriquesimoes (changes requested):
My review covers mainly changes to reduce the diff and to better organize the commits.

base/docker-compose.yml
```diff
@@ -28,4 +28,4 @@ services:
         MOTOR_VERSION: ${MOTOR_VERSION}
         PMAC_VERSION: ${PMAC_VERSION}
         LIBSSCPIMEGA_VERSION: ${LIBSSCPIMEGA_VERSION}
-        RETOOLS_VERSION: ${RETOOLS_VERSION}
+        RETOOLS_VERSION: ${RETOOLS_VERSION}
```
henriquesimoes: Missing last EOL here.

base/docker-compose.yml
```diff
@@ -28,4 +28,6 @@ services:
         MOTOR_VERSION: ${MOTOR_VERSION}
         PMAC_VERSION: ${PMAC_VERSION}
         LIBSSCPIMEGA_VERSION: ${LIBSSCPIMEGA_VERSION}
-        RETOOLS_VERSION: ${RETOOLS_VERSION}
\ No newline at end of file
+        RETOOLS_VERSION: ${RETOOLS_VERSION}
+        OPCUA_VERSION: ${OPCUA_VERSION}
+        ETHER_IP_VERSION: ${ETHER_IP_VERSION}
```
henriquesimoes, line 33: Same here.

base/musl/docker-compose.yml
```diff
@@ -24,3 +24,4 @@ services:
         IPAC_VERSION: ${IPAC_VERSION}
         CAPUTLOG_VERSION: ${CAPUTLOG_VERSION}
         RETOOLS_VERSION: ${RETOOLS_VERSION}
+        ETHER_IP_VERSION: ${ETHER_IP_VERSION}
```
henriquesimoes, line 27: Same here.

henriquesimoes (approved):
Everything looks fine by me. @ericonr, can you finish your review on this for us to merge it?

Highlighted some points that can be fixed when merging. No need for another patch iteration for these.

ericonr (approved):

Issue #24, opened 2023-10-27
Add label to the resulting image containing the build properties


ericonr:
At the very least, we should include the version of epics-in-docker that was used. If possible, we could try including information like the target that was used (e.g. `dynamic-link`).

PR #59, opened 2024-04-16
Prune unused artifacts from non-static builds

henriquesimoes:
Both `dynamic-link` and `no-build` targets generate unnecessarily large IOC images due to the trivial `COPY` done from `/opt`. Clean up everything from there and `/usr/local` directory before doing the `COPY` to the IOC image.

Clean up strategy proposed here is the following:

- Remove entirely any module that does not have a shared library linked against binaries under the IOC target directories (i.e. `/opt/$REPONAME` or `$RUNDIR`);
- Prune directories from used modules and EPICS base not containing shared libraries, EPICS databases or autosave requirement files;
- Remove any shared library not linked against binaries under IOC target directories;
- Remove all statically linked libraries;

For pruning module directories, another approach that could be followed is to delegate this responsibility to `install_module`. This way, we can have contextualized pruning for modules. On the other hand, this cannot be as aggressive as implemented in this proposal, as build-dependent files (the ones defining the build-system itself, for instance) must exist until the IOC build is complete.

I have tested this with [ad-aravis-epics-ioc](https://github.com/cnpem/ad-aravis-epics-ioc), and it shrinks the image size from 1.33GB down to 322MB.

This should be further tested and extensively reviewed, so that no artifact is removed unexpectedly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,8 +51,24 @@ services:
 
 By default, the IOC will be built with statically linked EPICS libraries. If
 you **need** to link them dynamically, you must define the build target as
-`dynamic-link`. This will increase the resulting image size, since unused
-dependencies will also be copied.
+`dynamic-link`.
+
+For non-static builds, an automatic pruning step is executed to make the IOC
+image size smaller by removing unused artifacts. This step preserves all ELF
+executables inside `RUNDIR` and the IOC repository, as well as their dependent
+EPICS modules based on the linkage information. Source code, GUI files, and
+other directories in used modules are assumed not to be needed at runtime and
+also removed. Additional paths can be provided in `args` under the `APP_DIRS`
+key to extend the list of where used applications and shared libraries are,
+including any `dlopen(3)`ed libraries. If the resulting image fails at runtime,
+a careful look at error messages might provide clues for directories to add to
+`APP_DIRS`; if that isn't possible, or if it's believed that the directories
+added to `APP_DIRS` should have been detected automatically, please open an
+issue. Please do so as well if the pruning step errors out during the build. In
+case it is necessary to build a working image before these issues can be fixed,
+the pruning step can be skipped entirely by setting the `SKIP_PRUNE=1`
+argument; note that this is highly discouraged, because it increases the image
+size significantly.
 
 The resulting image contains a standard IOC run script, `lnls-run`, which will
 be run inside `RUNDIR` and will launch the container's command under procServ,
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -35,7 +35,6 @@ RUN apt update -y && \
         ca-certificates
 
 COPY lnls-get-n-unpack.sh /usr/local/bin/lnls-get-n-unpack
-COPY lnls-run.sh /usr/local/bin/lnls-run
 
 ENV EPICS_IN_DOCKER=/opt/epics-in-docker
 RUN mkdir $EPICS_IN_DOCKER
@@ -68,3 +67,5 @@ RUN $EPICS_IN_DOCKER/install_motor.sh
 ARG DEBIAN_VERSION
 COPY opcua_versions.sh install_opcua.sh $EPICS_IN_DOCKER
 RUN $EPICS_IN_DOCKER/install_opcua.sh
+
+COPY lnls-run.sh /usr/local/bin/lnls-run
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -68,4 +68,5 @@ ARG DEBIAN_VERSION
 COPY opcua_versions.sh install_opcua.sh $EPICS_IN_DOCKER
 RUN $EPICS_IN_DOCKER/install_opcua.sh
 
+COPY lnls-prune-artifacts.sh /usr/local/bin/lnls-prune-artifacts
 COPY lnls-run.sh /usr/local/bin/lnls-run
--- a/base/lnls-prune-artifacts.sh
+++ b/base/lnls-prune-artifacts.sh
@@ -1,0 +1,113 @@
+#!/usr/bin/env bash
+
+set -Eeu
+
+# Filter out from the $1 list of paths any exact match of a parent directory
+# from any path in the $2 exclude list.
+#
+# Both lists are treated as newline-separated strings, and must consist
+# exclusively of absolute paths (except for empty lines which are ignored).
+filter_out_paths() {
+    list=$(echo "$1" | grep -v '^$')
+    exclude_list=$(echo "$2" | grep -v '^$')
+
+    while read -r path; do
+        if [ "${path:0:1}" != "/" ]; then
+            >&2 echo "error: filter_out_paths() expects absolute paths, but got '$path'"
+            exit 1
+        fi
+
+        while [ "$path" != "/" ]; do
+            list=$(echo "$list" | grep -xv "$path")
+
+            path=$(dirname "$path")
+        done
+    done <<< "$exclude_list"
+
+    echo "$list"
+}
+
+# Output the set $1 difference to another set $2
+compute_set_difference() {
+    sort $1 $2 $2 | uniq -u
+}
+
+find_elf_executables() {
+    targets=$@
+
+    # Loop on entire lines to properly handle filenames with spaces
+    while read -r file; do
+        read -r -N 4 magic < "$file"
+
+        # Output only ELF binaries
+        if [ "$magic" = $'\x7fELF' ]; then
+            echo $file
+        fi
+    done < <(find $targets -type f)
+}
+
+find_linked_libraries() {
+    executables=$(find_elf_executables $@)
+
+    # Depend on the glibc-specific behavior of supporting multiple executables
+    # to be queried at once
+    linked=$(ldd $executables 2>/dev/null | grep '=>')
+
+    # We grep out not found libraries, since they cannot be kept if we don't
+    # know where they are.
+    #
+    # Final binary may be actually runnable, since rpath of another binary may
+    # pull those not-found libraries
+    found="$(echo "$linked" | grep -v "not found")"
+
+    # Get their full path
+    libs=$(echo "$found" | cut -d' ' -f 3)
+
+    echo "$libs" | sort -u
+}
+
+get_all_epics_modules() {
+    release_defs=$(grep = ${EPICS_RELEASE_FILE} | cut -d'=' -f 2)
+
+    echo "$release_defs" | grep $EPICS_MODULES_PATH
+}
+
+get_used_epics_modules() {
+    linked_libs=$(find_linked_libraries $@)
+    all_modules=$(get_all_epics_modules)
+
+    unused_modules=$(filter_out_paths "$all_modules" "$linked_libs")
+
+    compute_set_difference <(echo "$all_modules") <(echo "$unused_modules")
+}
+
+remove_unused_epics_modules() {
+    targets=$(echo $@ | sed -E "s|\s+|\n|g")
+
+    all_modules=$(get_all_epics_modules)
+    used_modules=$(get_used_epics_modules $targets)
+    cat << EOF
+===
+Used EPICS modules
+===
+$used_modules
+===
+EOF
+
+    keep_paths=$(printf "$targets\n$used_modules")
+    unused_modules=$(filter_out_paths "$all_modules" "$keep_paths")
+
+    # Assume module paths do not contain spaces, as we mostly control them
+    for module in $unused_modules; do
+        # if we already removed it because of its top-level repository or
+        # because it is an IOC, move on to the next.
+        [ ! -d $module ] && continue
+
+        size=$(du -hs $module | cut -f 1)
+
+        echo "Removing module '$module' ($size)..."
+        rm -rf $module
+    done
+}
+
+remove_unused_epics_modules $@
--- a/base/lnls-prune-artifacts.sh
+++ b/base/lnls-prune-artifacts.sh
@@ -46,6 +46,12 @@ find_elf_executables() {
     done < <(find $targets -type f)
 }
 
+find_shared_libraries() {
+    elf_files=$(find_elf_executables $@)
+
+    echo "$elf_files" | grep -E "\.so(.[0-9]+)*$" | sort -u
+}
+
 find_linked_libraries() {
     executables=$(find_elf_executables $@)
 
@@ -151,5 +157,28 @@ remove_static_libraries() {
     done
 }
 
+remove_unused_shared_libraries() {
+    target_libs=$(find_shared_libraries $@)
+    linked_libs=$(find_linked_libraries $@)
+    remove_libs=$(find_shared_libraries /opt /usr/local)
+
+    keep_paths="$target_libs $linked_libs"
+    keep_paths="$keep_paths $(get_defined_paths_to_keep $remove_libs)"
+
+    for lib in $remove_libs; do
+        size=$(du -hs $lib | cut -f 1)
+
+        # if library is not found inside any $keep_paths, remove it
+        if find $keep_paths -path "$lib" -exec false {} + 2> /dev/null; then
+            echo "Removing shared library '$lib' ($size)..."
+
+            rm -f ${lib%.so*}.so*
+        else
+            echo "Keeping shared library '$lib' ($size)..."
+        fi
+    done
+}
+
 remove_unused_epics_modules $@
 remove_static_libraries /opt /usr/local
+remove_unused_shared_libraries $@
--- a/base/lnls-prune-artifacts.sh
+++ b/base/lnls-prune-artifacts.sh
@@ -81,6 +81,34 @@ get_used_epics_modules() {
     compute_set_difference <(echo "$all_modules") <(echo "$unused_modules")
 }
 
+# Traverse ancestor directories of each provided path, and concatenate all
+# their .lnls-keep-paths defined entries as absolute paths.
+get_defined_paths_to_keep() {
+    for path; do
+        if [ "${path:0:1}" != "/" ]; then
+            >&2 echo "error: get_defined_paths_to_keep() expects absolute paths, but got '$path'"
+            exit 1
+        fi
+
+        while true; do
+            keep_path_file="$path/.lnls-keep-paths"
+
+            if [ -f "$keep_path_file" ]; then
+                keep_paths=$(cat "$keep_path_file")
+
+                for keep_path in $keep_paths; do
+                    # output it as an absolute path
+                    realpath "$path"/$keep_path
+                done
+            fi
+
+            [ "$path" == "/" ] && break
+
+            path=$(dirname $path)
+        done
+    done | sort -u
+}
+
 remove_unused_epics_modules() {
     targets=$(echo $@ | sed -E "s|\s+|\n|g")
 
--- a/base/lnls-prune-artifacts.sh
+++ b/base/lnls-prune-artifacts.sh
@@ -110,4 +110,18 @@ EOF
     done
 }
 
+remove_static_libraries() {
+    for target; do
+        libs=$(find $target -type f -name *.a)
+
+        if [ -n "$libs" ]; then
+            size=$(du -hsc $libs | tail -n 1 | cut -f 1)
+
+            echo "Removing static libraries from $target ($size)"
+            rm -f $libs
+        fi
+    done
+}
+
 remove_unused_epics_modules $@
+remove_static_libraries /opt /usr/local
--- a/base/lnls-prune-artifacts.sh
+++ b/base/lnls-prune-artifacts.sh
@@ -76,6 +76,8 @@ get_all_epics_modules() {
     release_defs=$(grep = ${EPICS_RELEASE_FILE} | cut -d'=' -f 2)
 
     echo "$release_defs" | grep $EPICS_MODULES_PATH
+
+    echo $EPICS_BASE_PATH
 }
 
 get_used_epics_modules() {
--- a/base/lnls-prune-artifacts.sh
+++ b/base/lnls-prune-artifacts.sh
@@ -115,7 +115,39 @@ get_defined_paths_to_keep() {
     done | sort -u
 }
 
-remove_unused_epics_modules() {
+prune_directories() {
+    local targets="$1"
+    local keep_paths="$2"
+
+    remove_dirs=$(filter_out_paths "$targets" "$keep_paths")
+
+    while read -r remove_dir; do
+        # if we already removed it because of its parent directory, move on to
+        # the next.
+        [ ! -d "$remove_dir" ] && continue
+
+        size=$(du -hs "$remove_dir" | cut -f 1)
+
+        echo "Removing directory '$remove_dir' ($size)..."
+        rm -rf "$remove_dir"
+    done <<< "$remove_dirs"
+}
+
+prune_module_directories() {
+    module=$1
+
+    module_dirs=$(find $module -type d)
+    keep_paths=$(cat << EOF
+$(find_shared_libraries $module)
+$(find $module -type f -regex ".*\.\(cmd\|db\|template\|req\|substitutions\)" -printf "%h\n" | sort -u)
+$(get_defined_paths_to_keep $module)
+EOF
+    )
+
+    prune_directories "$module_dirs" "$keep_paths"
+}
+
+clean_up_epics_modules() {
     targets=$(echo $@ | sed -E "s|\s+|\n|g")
 
     all_modules=$(get_all_epics_modules)
@@ -129,18 +161,15 @@ $used_modules
 EOF
 
     keep_paths=$(printf "$targets\n$used_modules")
-    unused_modules=$(filter_out_paths "$all_modules" "$keep_paths")
+    prune_directories "$all_modules" "$keep_paths"
 
-    # Assume module paths do not contain spaces, as we mostly control them
-    for module in $unused_modules; do
-        # if we already removed it because of its top-level repository or
-        # because it is an IOC, move on to the next.
-        [ ! -d $module ] && continue
+    # Filter out targets to provide a way to disable module pruning in special
+    # cases
+    prune_dirs=$(filter_out_paths "$used_modules" "$targets")
 
-        size=$(du -hs $module | cut -f 1)
-
-        echo "Removing module '$module' ($size)..."
-        rm -rf $module
+    for dir in $prune_dirs; do
+        echo "Pruning module '$dir'..."
+        prune_module_directories $dir
     done
 }
 
@@ -179,6 +208,6 @@ remove_unused_shared_libraries() {
     done
 }
 
-remove_unused_epics_modules $@
+clean_up_epics_modules $@
 remove_static_libraries /opt /usr/local
 remove_unused_shared_libraries $@
--- a/images/docker-compose-mca.yml
+++ b/images/docker-compose-mca.yml
@@ -8,6 +8,6 @@ services:
       labels:
         org.opencontainers.image.source: https://github.com/cnpem/epics-in-docker
       args:
-        REPONAME: mca
+        APP_DIRS: /opt/epics/modules/mca
         RUNDIR: /opt/epics/modules/mca/iocBoot/iocAmptek
         RUNTIME_PACKAGES: libpcap0.8 libnet1 libusb-1.0-0
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,11 @@
 * base: update to Debian 12. by @ericonr in
   https://github.com/cnpem/epics-in-docker/pull/84
   * Refer to up to date README for new/updated `RUNTIME_PACKAGES`.
+* Prune unused artifacts from non-static builds by @henriquesimoes in
+  https://github.com/cnpem/epics-in-docker/pull/59
+  * Pruning will happen automatically, and most cases shouldn't require any
+    configuration. Refer to the README for instructions on how to configure the
+    procedure when needed.
 
 ## v0.12.0
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,9 +42,17 @@ RUN ln -s ${ENTRYPOINT} ./entrypoint
 ENTRYPOINT ["./entrypoint"]
 
 
+FROM build-image AS pruned-build
+
+ARG APP_DIRS
+ARG RUNDIR
+
+RUN lnls-prune-artifacts ${APP_DIRS} ${RUNDIR}
+
+
 FROM base AS no-build
 
-COPY --from=build-image /opt /opt
+COPY --from=pruned-build /opt /opt
 
 
 FROM build-image AS build-stage
@@ -70,10 +78,13 @@ RUN rm -rf .git/
 FROM build-stage AS dynamic-build
 
 ARG JOBS=1
+ARG APP_DIRS
 ARG RUNDIR
 ARG SKIP_TESTS
 
 RUN make distclean && make -j ${JOBS} && make $([ "$SKIP_TESTS" != 1 ] && echo runtests) && make clean && make -C ${RUNDIR}
+
+RUN lnls-prune-artifacts ${APP_DIRS} ${PWD} ${RUNDIR}
 
 
 FROM base AS dynamic-link
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,8 +46,9 @@ FROM build-image AS pruned-build
 
 ARG APP_DIRS
 ARG RUNDIR
+ARG SKIP_PRUNE
 
-RUN lnls-prune-artifacts ${APP_DIRS} ${RUNDIR}
+RUN if [ "$SKIP_PRUNE" != 1 ]; then lnls-prune-artifacts ${APP_DIRS} ${RUNDIR}; fi
 
 
 FROM base AS no-build
@@ -81,10 +82,11 @@ ARG JOBS=1
 ARG APP_DIRS
 ARG RUNDIR
 ARG SKIP_TESTS
+ARG SKIP_PRUNE
 
 RUN make distclean && make -j ${JOBS} && make $([ "$SKIP_TESTS" != 1 ] && echo runtests) && make clean && make -C ${RUNDIR}
 
-RUN lnls-prune-artifacts ${APP_DIRS} ${PWD} ${RUNDIR}
+RUN if [ "$SKIP_PRUNE" != 1 ]; then lnls-prune-artifacts ${APP_DIRS} ${PWD} ${RUNDIR}; fi
 
 
 FROM base AS dynamic-link
--- a/base/install_area_detector.sh
+++ b/base/install_area_detector.sh
@@ -19,6 +19,8 @@ git submodule update --init --depth 1 -j ${JOBS} \
 
 rm -rf .git
 
+echo 'ADSupport/lib/linux*/libHDF5*plugin.so' > .lnls-keep-paths
+
 cd configure
 
 module_releases="
--- a/base/install_modules.sh
+++ b/base/install_modules.sh
@@ -96,6 +96,7 @@ echo PYTHON=python3 >> pyDevSup/configure/CONFIG_SITE
 install_module pyDevSup PYDEVSUP "
 EPICS_BASE
 "
+echo 'python3*/linux*/' > pyDevSup/.lnls-keep-paths
 
 mkdir snmp
 cd snmp
--- a/images/docker-compose-motorpigcs2.yml
+++ b/images/docker-compose-motorpigcs2.yml
@@ -8,5 +8,5 @@ services:
       labels:
         org.opencontainers.image.source: https://github.com/cnpem/epics-in-docker
       args:
-        REPONAME: motorpigcs2
+        APP_DIRS: /opt/epics/modules/motor/modules/motorPIGCS2/iocs/pigcs2IOC
         RUNDIR: /opt/epics/modules/motor/modules/motorPIGCS2/iocs/pigcs2IOC/iocBoot/iocPIGCS2
--- a/images/docker-compose-opcua.yml
+++ b/images/docker-compose-opcua.yml
@@ -8,6 +8,6 @@ services:
       labels:
         org.opencontainers.image.source: https://github.com/cnpem/epics-in-docker
       args:
-        REPONAME: opcua
+        APP_DIRS: /opt/epics/modules/opcua
         RUNDIR: /opt/epics/modules/opcua/iocBoot/iocUaDemoServer
         RUNTIME_PACKAGES: libxml2 libssl3
--- a/images/docker-compose-pvagw.yml
+++ b/images/docker-compose-pvagw.yml
@@ -8,5 +8,6 @@ services:
       labels:
         org.opencontainers.image.source: https://github.com/cnpem/epics-in-docker
       args:
+        APP_DIRS: /opt/epics/modules/p4p
         RUNDIR: /opt/epics/modules/p4p/bin/linux-x86_64
         RUNTIME_PACKAGES: python3-numpy python3-ply
--- a/images/docker-compose-pvagw.yml
+++ b/images/docker-compose-pvagw.yml
@@ -10,4 +10,5 @@ services:
       args:
         APP_DIRS: /opt/epics/modules/p4p
         RUNDIR: /opt/epics/modules/p4p/bin/linux-x86_64
-        RUNTIME_PACKAGES: python3-numpy python3-ply
+        ENTRYPOINT: ./pvagw
+        RUNTIME_PACKAGES: python3-numpy python3-ply libevent-pthreads-2.1-7


base/Dockerfile
```diff
@@ -66,3 +66,5 @@ RUN ./install_area_detector.sh
 
 COPY install_motor.sh .
 RUN ./install_motor.sh
+
+COPY lnls-prune-artifacts.sh /usr/local/bin/lnls-prune-artifacts
```
henriquesimoes, line 70: I've added this last so that image cache is not invalided everything for nothing. It can be moved to be side-by-side with `lnls-run` after the script ready to be merged.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,122 @@
...
+}
+
+get_all_epics_modules() {
+    release_defs=$(cat ${EPICS_RELEASE_FILE} | grep = | cut -d'=' -f 2)
+
+    modules="$(echo "$release_defs" | grep $EPICS_MODULES_PATH)"
+
+    for module in $modules; do
+        echo $module
+    done
```
henriquesimoes, line 47: Doing this instead of simply echoing out `"$modules"` removes empty lines. Maybe there is a shorter and cleaner way to do this.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,122 @@
...
+}
+
+remove_static_libs() {
+    for target; do
+        libs=$(find $target -type f -name *.a -print)
+
+        if [ -n "$libs" ]; then
+            size=$(du -hsc $libs | tail -n 1 | cut -f 1)
+
+            echo "Removing static libraries from $target ($size)"
```
henriquesimoes, line 57: I'm collapsing all static libraries. Should we list the removed files instead?

I decided to let it collapsed because it is simpler to implement and no application should need them after build.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,122 @@
...
+    modules="$(echo "$release_defs" | grep $EPICS_MODULES_PATH)"
+
+    for module in $modules; do
+        echo $module
+    done
+}
+
+remove_static_libs() {
+    for target; do
+        libs=$(find $target -type f -name *.a -print)
```
henriquesimoes, line 52: I'm assuming there are only [AR archives](https://en.wikipedia.org/wiki/Ar_(Unix)) which are statically linked libraries. Is that a really good assumption?

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,122 @@
...
+    for lib in $used_libs; do
+        remove_libs=$(echo "$remove_libs" | grep -v $lib)
+    done
+
+    for lib in $remove_libs; do
+        size=$(du -hs $lib | cut -f 1)
+
+        echo "Removing shared library '$lib' ($size)"
+        rm -f $lib
+    done
```
henriquesimoes, line 78: I don't know if there is an easier way to do the set difference besides looping twice while `grep`ing out.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,122 @@
...
+
+    for candidate in $(find $module -type d); do
+        [ -d $candidate ] || continue
+
+        if [[ ! $keep_paths =~ "$candidate".* ]]; then
+            size=$(du -hs $candidate | cut -f 1)
+
+            printf "Removing directory '$candidate' ($size)...\n"
+            rm -rf $candidate
+        fi
```
henriquesimoes, line 97: In this loop, I'm removing only directories as a whole if no important file (listed in `keep_paths`) is stored in it.

Another approach is to do this in a file basis, which could prune modules even further. I'm not sure if that's the best way to handle it, though.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,122 @@
...
+
+    for module in $(get_all_epics_modules); do
+        if [[ -d $module && ! $used_modules =~ "$module" ]]; then
+            size=$(du -hs $module | cut -f 1)
+
+            echo "Removing module '$module' ($size)..."
+            rm -rf $module
+        elif [ -d $module/lib ]; then
+            prune_module_dirs $module
+        fi
```
henriquesimoes, line 114: I'm using the `lib` subdirectory to ignore top-level repositories listed in `RELEASE`, such as areaDetector. Is there a better approach?

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,122 @@
...
+    done
+
+    prune_module_dirs $EPICS_BASE_PATH
+}
+
+clean_up_epics_modules $target_paths
+remove_static_libs /opt
+remove_unused_shared_libs $target_paths
```
henriquesimoes, line 122: I'm not sure if `remove_static_libs` should receive an argument or not. Their companion functions receive directories which contain important binaries whose dependencies must be preserved. On the other hand, `remove_static_libs` currently receives the directories to prune the libraries off.

Maybe both of them should take `/opt` (and `/usr/local`?) from a shared constant variable instead.

Opinions on that?

henriquesimoes (commented):
I've added comments to points that I wasn't sure what is the best way to deal with, so that we can discuss if there are relevant or not and how to improve, if that's the case.

In general, I feel like I could have written helper functions that could be more easily shared among the steps. But that's not a concrete thing to propose anything. Let me know if you feel the same.

base/lnls-prune-artifacts.sh
ericonr: I'd use the sed answer from <https://serverfault.com/questions/252921/how-to-remove-empty-blank-lines-from-a-file-in-unix-including-spaces>

base/lnls-prune-artifacts.sh
ericonr: I'm okay with them being collapsed.

base/lnls-prune-artifacts.sh
ericonr: For running on Linux, I think it should be.

If you want to have a sort of lint for our build process, after the base image is done building stuff, I'd run a script to make sure we don't have any `.o` files leftover. They are the only other kind of "static" files we could have.

base/lnls-prune-artifacts.sh
ericonr: > I don't know if there is an easier way to do the set difference

This was the part that made me question if python wouldn't be best. (and calling `ldd` as a subprocess in python would be ok for me, with the added advantage of python allowing more expressiveness in text processing).

base/lnls-prune-artifacts.sh
ericonr: I can think of hardcoding things, but that doesn't really seem better. Will think on it.

base/lnls-prune-artifacts.sh
henriquesimoes: I've checked our base image, and it does not have any `*.o` left behind. Our cleaning done after installation seems to be enough for now to remove these.

ericonr (commented):
The first two commits are correct and can be merged separately, if you'd like to split them.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+
+set -eu
+
+target_paths=$@
+
+
```
ericonr: I don't think this variable needs to be set; the function call at the end can simply use `$@` itself.

Putting this at the start of the script makes it look like you are going to manipulate the array in the middle of it.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,45 @@
...
+get_used_epics_modules() {
+    targets=$@
+
+    executables=$(find $targets -type f -executable -print)
```
ericonr, line 29: `-print` is the default, and I had to search for it to see if it changed anything and confirm it's the default. I'd rather omit it.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,45 @@
...
+    libs="$(ldd $executables 2>/dev/null)"
+
+    modules=()
```
ericonr, line 32: Leftover variable

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,45 @@
...
+set -eu
+
+target_paths=$@
+
+
+get_used_epics_modules() {
+    targets=$@
+
+    executables=$(find $targets -type f -executable -print)
+    libs="$(ldd $executables 2>/dev/null)"
```
ericonr, line 30: `ldd` on musl doesn't check all arguments, it simply ignores all but the first one.

We should either explicitly loop or add a comment about depending on this behavior.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,45 @@
...
+
+    modules=()
+    epics_libs=$(echo "$libs" | grep -oE "$EPICS_MODULES_PATH/.*/*.so")
```
ericonr, line 33: I don't think the EPICS build system supports versioned sonames, so using `.so` is fine. Maybe mention that in the commit message?

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,45 @@
...
+    for lib in $epics_libs; do
+        module=${lib%/lib/*/*lib*so}
```
ericonr, line 35: Use `epics_lib` or even something other than `lib`. Otherwise the substitution gets confusing with the repeated `lib`, and it's confusing enough already.

Add a comment explaining the purpose of expression (an example would be enough).

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,45 @@
...
+        echo $module
+    done | sort | uniq
```
ericonr, line 37: `sort -u` so we don't need `uniq`

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,45 @@
...
+}
+
+get_all_epics_modules() {
+    release_defs=$(cat ${EPICS_RELEASE_FILE} | grep = | cut -d'=' -f 2)
```
ericonr, line 41: `grep = ${EPICS_RELEASE_FILE}`

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,45 @@
...
+
+    modules="$(echo "$release_defs" | grep $EPICS_MODULES_PATH)"
+
+    echo "$modules"
```
ericonr, line 45: You could simply have the last line be `echo $release_defs | grep ...`

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,45 @@
...
+}
+
+remove_unused_epics_modules() {
+    targets=$@
+
+    used_modules=$(get_used_epics_modules $targets)
+
+    for module in $(get_all_epics_modules); do
+        if [[ -d $module && ! $used_modules =~ "$module" ]]; then
+            size=$(du -hs $module | cut -f 1)
```
ericonr, line 106: Why is `-d $module` required? Shouldn't they always exist? Or is this covering some case like an AreaDetector or motor submodule being in the list after the super-repository has already been cleaned? I think this should be explained somewhere.

I find `=~` somewhat non obvious to parse. And it means "motor" won't be removed if "motorX" is in used modules, right? I think it might make sense to print the list into a file (in /tmp, which can be removed afterwards), so it's split by newlines, and use grep to match the *whole* line.

ericonr (commented):
The Dockerfile parts of the third commit are ok.

---

Commit message nits:

It's not the caching that's the issue, it's simply how container images are layered:

```
The clean up phase must be executed in the build stage (before the final
copy), otherwise the COPY layer would still make the image size big.
This implies `no-build` target needs to copy from a pruned version of
the base image, which is a new stage.
```

base/lnls-prune-artifacts.sh
henriquesimoes: Let me know what you think about the comment.

base/lnls-prune-artifacts.sh
henriquesimoes: Such hack is no longer needed with the latest rewrite. I'm filtering out target directories (and their parents), so we shouldn't call for top-level repos.

base/lnls-prune-artifacts.sh
henriquesimoes: > And it means "motor" won't be removed if "motorX" is in used modules

Because of such issues with prefix matching, I moved to checking the each level of the path at a time, with whole line matching instead. This turned out to be handy for the following patches as well.

Still, don't feel like my implementation explores all bash-ims it could.

images/docker-compose-mca.yml
```diff
@@ -8,6 +8,6 @@ services:
       labels:
         org.opencontainers.image.source: https://github.com/cnpem/epics-in-docker
       args:
-        REPONAME: mca
+        REPONAME: epics/modules/mca
```
henriquesimoes, line 11: To be honest, I don't like very much the idea of reusing `REPONAME`. It seems awkward to me. I'd rather have them specified by a variable whose name makes sense at all.

I did so because it was easier to test other things first. Let me know what you think.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,91 @@
...
+    for module in $(filter_out_dirs "$unused_modules" "$keep_dirs"); do
+        # if we already removed it because of its top-level repository or
+        # because it is an IOC, move on to the next.
+        [ ! -d $module ] && continue
+
+        size=$(du -hs $module | cut -f 1)
+
+        echo "Removing module '$module' ($size)..."
+        rm -rf $module
+    done
```
henriquesimoes, line 143: This loop will misbehave if we turn out having modules with empty spaces.

I kept this as a loop with such requirement because we mostly control such paths. Let me know what you think about this.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,91 @@
...
+    targets=$@
+
+    while read -r executable; do
+        read -r -N 4 trailer < "$executable"
+
+        # Output only ELF binaries
+        if [ "$trailer" = $'\x7fELF' ]; then
+            echo $executable
+        fi
+    done < <(find $targets -type f -executable)
```
henriquesimoes, line 33: We must loop this way here because they might be file names with spaces. Indeed, MCA is such a representative of such scenario.

Because of that, I wonder if I haven't assumed to much elsewhere regarding file names.

Also, would this deserve a comment or is this clear enough?

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,91 @@
...
+
+    # We grep out not found libraries, since they cannot be kept if we don't
+    # know where they are.
+    #
+    # Final binary may be actually runnable, since rpath of another binary may
+    # pull those not found libraries
+    found="$(echo "$linked" | grep -v "not found")"
+
+    # Get their full path
+    libs=$(echo "$found"| cut -d'>' -f 2 | cut -d' ' -f 2)
```
henriquesimoes, line 57: I feel like there is a better way to do this. I though abort `grep -o`, but it would require a pattern for any library path, including soname handling. I'm not sure if that turns out to be better.

images/docker-compose-mca.yml
ericonr: That's fair. Do you think it would lead to special casing these modules in the image Dockerfile?

henriquesimoes: > Do you think it would lead to special casing these modules in the image Dockerfile?

They kinda already have such special handling... all of them are no-build targets. They of course currently have almost the same (compose variables) API of other targets. I say "almost" because they diverged by not using REPONAME before, and now by not meaning "base name to where to clone the repository" but rather "location inside /opt where important artifacts live".

I'm not sure if I answered your question at all, but I do feel like we have always being dealing with a special case.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,91 @@
...
+get_all_epics_modules() {
+    release_defs=$(grep = ${EPICS_RELEASE_FILE} | cut -d'=' -f 2)
+
+    echo "$release_defs" | grep $EPICS_MODULES_PATH
+}
+
+remove_unused_epics_modules() {
+    targets=$@
+
+    unused_modules=$(get_all_epics_modules)
```
ericonr, line 129: This variable has the wrong name.

ericonr (changes requested):
One thing isn't clear to me. From targets, you are obtaining unused modules, to then obtain used modules, to finally obtain unused modules again, and remove them. Is that logic necessary, can we simplify it? If you already have these answers, you can add them to the commit message.

Commit message nit:

> EPICS base is assumed to be always needed,

always be needed

> It is assumed we are using glibc implementation

we are using glibc's ldd

For the second commit, since we are removing stuff *after* building, static libraries can always be removed, notjust for non-static builds. I think it's only the commit message that's wrong.

Nit:

> Static libraries are not used neither during compile time nor runtime for non-static builds.

Static libraries are not used during compile nor runtime.

Static libraries are used neither during compile nor runtime.

> All AR archives are assumed to be static libraries,

All files with .a extension are assumed to be static libraries,

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,91 @@
...
+
+    # Get their full path
+    libs=$(echo "$found"| cut -d'>' -f 2 | cut -d' ' -f 2)
+
+    echo "$libs" | sort -u
+}
+
+get_used_epics_modules() {
+    epics_libs=$(find_linked_libraries $@)
+    modules=$(get_all_epics_modules)
```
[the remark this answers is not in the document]
henriquesimoes: Okay. I didn't take this into account when ordering the functions. Will reorder them. Thanks.

base/lnls-prune-artifacts.sh
henriquesimoes: > Do you mean using this instead of for executable in $(find ...?

Exactly.

> You can just leave a comment on top of the loop that file names might have spaces.

Okay. Should I do the same for every other loop as well?

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,91 @@
...
+    # Depend on the glibc-specific behavior of supporting multiple executables
+    # to be queried at once
+    linked=$(ldd $executables 2>/dev/null | grep '=>')
+
+    # We grep out not found libraries, since they cannot be kept if we don't
+    # know where they are.
+    #
+    # Final binary may be actually runnable, since rpath of another binary may
+    # pull those not found libraries
+    found="$(echo "$linked" | grep -v "not found")"
```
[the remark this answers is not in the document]
henriquesimoes: Yep. For `libopcua.so` that comes in OPCUA module, both EPICS libCom and libdbCore won't be found if we simply `ldd` it.

images/docker-compose-mca.yml
ericonr: I guess I am essentially wondering if we shouldn't use some other name in REPONAME, since these images are a special case.

base/lnls-prune-artifacts.sh
ericonr: > I kept this as a loop with such requirement because we mostly control such paths. Let me know what you think about this.

I agree with this justification here, especially because it makes loops more complicated.

base/lnls-prune-artifacts.sh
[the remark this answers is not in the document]
ericonr: > EPICS libCom and libdbCore won't be found if we simply ldd it.

Oh, that's annoying. Is that the only case?

I'm wondering if we should make it clear it's due to OPCUA, and remove it once/if we move to the open source impl, or if it makes sense to keep for any other similar cases with proprietary libraries.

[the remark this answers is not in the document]
henriquesimoes: > Is that the only case?

That's the only one that I spotted. Considering that EPICS modules and IOC by default relies on those custom rpath entries, I'm sure any other already-compiled library shipped into module is susceptible to this issue. This probably falls into what you mention of proprietary libraries in most cases.

If we were using system installation for EPICS base, that wouldn't have been an issue even for OPCUA.

base/lnls-prune-artifacts.sh
henriquesimoes: > Why doesn't using one `cut -d' ' -f 3` command work?

Actually, it does. I was trying not to rely on spaces at first, but I turned out doing so in the end. I didn't notice it could be easily simplified. Thanks.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,91 @@
...
+    exclude_list="$2"
+
+    for item in $exclude_list; do
+        dir=$item
+
+        while [ "$dir" != "/" ]; do
+            list=$(echo "$list" | grep -xv $dir)
+
+            dir=$(dirname $dir)
+        done
```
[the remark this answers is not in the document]
henriquesimoes: Yeah, I turned out to do so when testing to avoid infinite loops, but didn't get to a final approach on how to handle it.

Should we abort in such cases, or simply ignore that entry?

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,91 @@
...
+
+set -eu
+
+
+filter_out_dirs() {
+    list="$1"
+    exclude_list="$2"
+
+    for item in $exclude_list; do
+        dir=$item
```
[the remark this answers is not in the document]
henriquesimoes: > This should include a comment indicating that for the first iteration, dir might not actually be a directory.

Should we call it `path` instead?

> And why not use for dir in...?

Good question. None, I'm afraid. When expressing the logic, that was the natural way to implement it, and I didn't simplify it in the end. Thanks for the notice.

base/lnls-prune-artifacts.sh
henriquesimoes: Oh, thanks. Refactoring left over mistake for sure. `all_modules` should do the job.

base/lnls-prune-artifacts.sh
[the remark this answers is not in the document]
ericonr: IMO abort, it's breaking an expected invariant.

base/lnls-prune-artifacts.sh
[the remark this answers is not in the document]
ericonr: > Should we call it path instead?

I think that works well!

base/lnls-prune-artifacts.sh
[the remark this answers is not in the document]
ericonr: I'm pending towards mentioning in a comment that the "not-found" case is due to OPCUA, then.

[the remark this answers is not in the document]
henriquesimoes: Added to the commit message. Let me know if you think that's not enough.

images/docker-compose-mca.yml
henriquesimoes: I changed this to `APP_DIRS`. It should encode full paths now. It is named after a plural noun; after all, we can have multiple directories that we want to ensure are properly working. This should work as a palliative measure for spurious cases where we clean up too much. Any directory listed there will have its belonging module removed from pruning (both module-based and directory-based).

Maybe we could be even more explicit about its usage regarding "not pruning it too much and keeping its executable binaries working". Any opinion and suggestion about that?

gustavosr8:
nit from 13072def851f1ea2ff39b96aca02b86f840da9e6:
`s/clean-up/cleanup`
`s/clean up/cleanup`
`s/so that it contains a valid mapping of all modules/so it contains a valid mapping of all modules` 
`s/assumes the inspected binaries/assumes that the inspected binaries`
`s/It is assumed we are using glibc's/It is assumed that we are using glibc's`

gustavosr8:
> Thanks, @gustavosr8. I've also made `remove_unused_shared_libs` respect libraries inside `APP_DIRS`. This allows users to specify the path where their `dlopen(3)`ed libraries live. Can you test this again with the pyDevSup-based IOC?

Now it seems to work! Thanks :)
Just a reminder to document the use of `APP_DIRS` ;)

gustavosr8 (changes requested):
Please add instructions to put `/opt/epics/modules/pyDevSup/python3.9` under `APP_DIRS` (or any analogous var) when building an IOC that uses the pyDevSup module.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env bash
+
+set -Eeu
```
ericonr, line 3: Mention in commit message that you need `-E` to properly propagate errors.

images/docker-compose-mca.yml
```diff
@@ -8,6 +8,6 @@ services:
       labels:
         org.opencontainers.image.source: https://github.com/cnpem/epics-in-docker
       args:
-        REPONAME: mca
+        APP_DIRS: /opt/epics/modules/mca
```
ericonr, line 11: Also mention in commit message that REPONAME wasn't relevant for these images, and that they were moved to use `APP_DIRS` so as not to break them.

base/lnls-prune-artifacts.sh
```diff
@@ -106,5 +112,23 @@ remove_static_libs() {
     done
 }
 
+remove_unused_shared_libs() {
+    target_libs=$(find_shared_libs $@)
+    linked_libs=$(find_linked_libraries $@)
+    remove_libs=$(find_shared_libs /opt /usr/local)
```
ericonr, line 171: Adding `/usr/local` here is due to #55, right? Might make sense to mention in the commit that it's future proofing, and doesn't actually change behavior for now.

base/lnls-prune-artifacts.sh
```diff
@@ -36,6 +36,12 @@ find_elf_executables() {
     done < <(find $targets -type f -executable)
 }
 
+find_shared_libs() {
```
ericonr, line 39: Consistency: libraries vs libs

base/lnls-prune-artifacts.sh
```diff
@@ -77,7 +77,27 @@ get_used_epics_modules() {
     filter_out_paths "$all_modules" "$unused_modules"
 }
 
-remove_unused_epics_modules() {
+prune_module_dirs() {
+    module=$1
+
+    keep_paths="
+        $(find_shared_libs $module)
+        $(find $module -type f -regex ".*\.\(db\|template\|req\)" -printf "%h\n" | sort -u)
```
ericonr, line 108: Some IOCs use `substitutions` at runtime, I think they should also be kept...

Should ACFs be kept as well?

base/lnls-prune-artifacts.sh
```diff
@@ -77,7 +77,27 @@ get_used_epics_modules() {
...
+
+    while read -r candidate; do
+        [ -d $candidate ] || continue
+
+        if [[ ! $keep_paths =~ "$candidate".* ]]; then
+            size=$(du -hs $candidate | cut -f 1)
+
+            printf "Removing directory '$candidate' ($size)...\n"
+            rm -rf $candidate
+        fi
```
ericonr, line 120: Is this nearly the same operation as `filter_out_paths`? Is it open-coded because you need to support spaces?

base/lnls-prune-artifacts.sh
```diff
@@ -77,12 +77,36 @@ get_used_epics_modules() {
...
 }
 
+# Traverse ancestor directories of each provided path, and concatenate all
+# their .lnls-keep-paths defined entries as absolute paths.
+get_defined_paths_to_keep() {
+    for path; do
+        depth=$(echo $path | grep -o '/' | wc -l)
+
+        for i in $(seq 1 $depth); do
+            ancestor=$(echo $path | cut -d '/' -f 1-$i)
```
ericonr, line 87: Why can't this be a `dirname` loop that terminates on `/`?

ericonr (commented):
Commit message nits:

"workaround" should be "work around", it's being used as a verb.

"a "import"" should use "an"

ericonr:
I'm force pushing into the PR, after rebasing it and fixing some of my own nits.

Original commit: 0a9c79e85dcc5344798a594583f2874c6f154000

base/lnls-prune-artifacts.sh
henriquesimoes: Yes, and yes. However, regex matching suffers from what you said elsewhere: if a candidate folder is a prefix of a path to keep, this method decides to keep them. This means we might remove less than we could.

Maybe supporting file paths with spaces to `filter_out_paths` is a way to go. What do you think?

ericonr: That would mean you need to use arrays instead of lists in strings, right?

base/lnls-prune-artifacts.sh
ericonr: Could also break on `.` and `/`, so it still works for local directories.

henriquesimoes: > I like the dirname loop because it's already used elsewhere, so it's a pattern that's easy to spot.

Yeah, I agree on that. I changed this locally. I'm testing if it is still working as expected before pushing.

> Could also break on . and /, so it still works for local directories.

Indeed. I'll defer that to the future if needed. As I said, it actually is not a big deal for now.

base/lnls-prune-artifacts.sh
henriquesimoes: Don't think so. `filter_out_paths` already receives two newline-separated strings. Stop using `for` and redirect the list to the `while read; do ... done` stdin should be enough.

henriquesimoes: Doing so made possible to extract yet another function to be used both when removing entire modules and pruning only specific directories. ;)

base/Dockerfile
henriquesimoes: Made this a permanent change now. ;)

henriquesimoes:
Changed the last implementation bits now. I learned something new during this round: calling a bash function outside a subshell does not create another variable scope. I was assuming it did until now. If I knew this from start, I would have defined several other variables explicitly as `local` just to avoid overwriting something by mistake in the future.

I haven't done so now, though, to reduce the amount of noise in range diffs. I've only done so now where it would be a bug otherwise. Let me know if you think it is interesting to refactor that before merging.

From my point of view, documentation is the last topic to be covered in this PR.

README.md
```diff
@@ -51,8 +51,17 @@ services:
...
+
+For non-static builds, an automatic pruning step is executed to make the IOC
+image size small by removing unused artifacts. This step preserves all ELF
+executables inside `RUNDIR` and the IOC repository, as well as their dependent
+EPICS modules based on the linkage information. Additional paths can be
+provided in `args` under `APP_DIRS` key to extend the list of where important
+applications and shared libraries are, including any `dlopen(3)`ed libraries.
+Pruning step can be skipped entirely by providing the `SKIP_PRUNE=1` argument,
+but this is highly discouraged as it will increase significantly the image
+size.
```
henriquesimoes, line 64: Do you think it is interesting to specify we also prune directories from modules in use?

I don't like the idea of replicating what types of files are kept, but it may be okay to say they are also pruned.

ericonr, line 64: If the resulting image fails at runtime, a careful look at error messages might provide clues for directories to add to `APP_DIRS`; if that isn't possible, or if it's believed that the directories added to `APP_DIRS` should have been detected automatically, please open an issue. Please do so as well if the pruning step errors out during the build. In case it is necessary to build a working image before these issues can be fixed, the pruning step can be skipped entirely by setting the `SKIP_PRUNE=1` argument; note that this is highly discouraged, because it increases the image size significantly.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,198 @@
+#!/usr/bin/env bash
+
+set -Eeu
+
+filter_out_paths() {
```
ericonr, line 5: This function is missing a comment that its arguments should be split by newlines

henriquesimoes (commented):
Two nits left I've noticed now:

- clean up `/opt/epics-in-docker` (#88)
- clean up `lnls-*` scripts we don't need (#59)

Maybe the last one can be deferred to the #59 itself, even though `lnls-get-n-unpack` cannot be removed by `lnls-prune-artifacts`.

ericonr (changes requested):

README.md
ericonr: I think it should be mentioned that some directories (e.g. GUI ones) are assumed to be unused and removed as well.

ericonr:
> Two nits left I've noticed now:
> 
>     * clean up `/opt/epics-in-docker` ([base: remove repetition of environment variables. #88](https://github.com/cnpem/epics-in-docker/pull/88))
> 
>     * clean up `lnls-*` scripts we don't need ([Prune unused artifacts from non-static builds #59](https://github.com/cnpem/epics-in-docker/pull/59))
> 
> 
> Maybe the last one can be deferred to the #59 itself, even though `lnls-get-n-unpack` cannot be removed by `lnls-prune-artifacts`.

is 59 there correct? that's this PR

I'm reconsidering cleaning up `/opt/epics-in-docker`. It's composed of small text files, and the info there is a pretty easy way to see what versions of everything were used to build a certain image.

henriquesimoes:
> is 59 there correct? that's this PR

Oops. It is not. It is #55.

> I'm reconsidering cleaning up /opt/epics-in-docker. It's composed of small text files, and the info there is a pretty easy way to see what versions of everything were used to build a certain image.

Makes sense. Based on the first point, maybe it is not worth to remove auxiliary scripts as well.

gustavosr8:
> I'm reconsidering cleaning up `/opt/epics-in-docker`. It's composed of small text files, and the info there is a pretty easy way to see what versions of everything were used to build a certain image.

Maybe if we use `LABELS` to keep track of these infos (#24) we could clear this up. WDYT?

ericonr:
> > I'm reconsidering cleaning up `/opt/epics-in-docker`. It's composed of small text files, and the info there is a pretty easy way to see what versions of everything were used to build a certain image.
> 
> Maybe if we use `LABELS` to keep track of these infos (#24) we could clear this up. WDYT?

I think I'd be okay with both rather than removing the files, either way.

README.md
henriquesimoes: Done. Feel free to propose an enhanced wording ;)

base/Dockerfile
```diff
@@ -35,7 +35,6 @@ RUN apt update -y && \
         ca-certificates
 
 COPY lnls-get-n-unpack.sh /usr/local/bin/lnls-get-n-unpack
-COPY lnls-run.sh /usr/local/bin/lnls-run
```
ericonr, line 38: This commit message is wrong. This script isn't executed during any build step.

ericonr (approved):

gustavosr8 (approved):

base/Dockerfile
henriquesimoes: Indeed. I was willing to write in the affirmative tense, but made the complement subset smaller than it actually is. Thanks.

henriquesimoes:
Something is off when building the opcua image: it removes [`lib/` from `opcua-module`](https://github.com/cnpem/epics-in-docker/pull/59/checks#step:8:448). I'll investigate that.

henriquesimoes:
Okay, I've found the issue: epics-modules/opcua#169. As discussed in person with @ericonr, I'll drop this assumption of libraries having the execution bit set. This will make we check the first bytes of all files inside a module to be pruned, but it should give us a more general approach to deal with any library installed (including the ones from non-EPICS modules).

For building `opcua-epics-ioc` in my local machine (5 times), pruning by looking only executable files took $1.71 \pm 0.05$ second while checking all files took $1.77 \pm 0.07$ second. For `ad-aravis-epics-ioc`, it took $5.1 \pm 0.2$ seconds (only exe) vs $5.25 \pm 0.09$ seconds (all files). In the end, It looks like that's basically for free.

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,101 @@
...
+    echo "$release_defs" | grep $EPICS_MODULES_PATH
+}
+
+get_used_epics_modules() {
+    linked_libs=$(find_linked_libraries $@)
+    all_modules=$(get_all_epics_modules)
+
+    unused_modules=$(filter_out_paths "$all_modules" "$linked_libs")
+
+    filter_out_paths "$all_modules" "$unused_modules"
```
henriquesimoes, line 84: This logic is actually broken. If we get in `$unused_modules` something including `/opt/epics/modules/motor/modules/motorMotorSim`, for instance, the second `filter_out_paths` will also remove `/opt/epics/modules/motor`, since it is a prefix of the first one. However, we don't want such "prefixes" removal in this last step, as it should be only a pure and simple set difference.

I'll fix that.

henriquesimoes:
I've just checked a couple of IOCs:

- our four built-in images;
- ad-aravis-epics-ioc
- ad-sim-epics-ioc
- motor-sim-epics-ioc

My tests consisted in building the container themselves, and checking if binaries run at all. I haven't checked all the files removal logs carefully for all of them, though. For the ones I did, everything seemed to make sense.

One thing I noticed during my tests with ad-sim-epics-ioc is that all directories from modules inside areaDetector are pruned at once when the top-level repo is pruned. This makes inner modules pruning basically a no-op. It doesn't cause anything harmful for this IOC, though. It does mean `.lnls-keep-paths` but live in `areaDetector` top-level directory (I now think I should have added this to the commit message, BTW).

<details>
<summary>Logs of pruning areaDetector modules for ad-sim-epics-ioc</summary>

```
Pruning module '/opt/epics/modules/areaDetector'...
Removing directory '/opt/epics/modules/areaDetector/ADCSimDetector' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADnED' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADBruker' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADPcoWin' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADRIXSCam' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADSupport/configure' (36K)...
Removing directory '/opt/epics/modules/areaDetector/ADSupport/supportApp' (94M)...
Removing directory '/opt/epics/modules/areaDetector/ADSupport/include' (1.2M)...
Removing directory '/opt/epics/modules/areaDetector/ADSupport/bin' (208K)...
Removing directory '/opt/epics/modules/areaDetector/ADUVC' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADProsilica' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ci' (16K)...
Removing directory '/opt/epics/modules/areaDetector/ADPluginBar' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADPSL' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADPhotonII' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADSpinnaker' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADPixirad' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/pvaDriver' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADPCO' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADPICam' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/configure' (136K)...
Removing directory '/opt/epics/modules/areaDetector/documentation' (148K)...
Removing directory '/opt/epics/modules/areaDetector/ADSimDetector/configure' (36K)...
Removing directory '/opt/epics/modules/areaDetector/ADSimDetector/documentation' (652K)...
Removing directory '/opt/epics/modules/areaDetector/ADSimDetector/include' (12K)...
Removing directory '/opt/epics/modules/areaDetector/ADSimDetector/iocs/simDetectorIOC/iocBoot/iocSimDetector/autosave' (8.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADSimDetector/iocs/simDetectorIOC/iocBoot/iocHDF5Test/autosave' (8.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADSimDetector/iocs/simDetectorIOC/configure' (36K)...
Removing directory '/opt/epics/modules/areaDetector/ADSimDetector/iocs/simDetectorIOC/simDetectorApp' (28K)...
Removing directory '/opt/epics/modules/areaDetector/ADSimDetector/iocs/simDetectorNoIOC' (68K)...
Removing directory '/opt/epics/modules/areaDetector/ADSimDetector/simDetectorApp/src' (68K)...
Removing directory '/opt/epics/modules/areaDetector/ADSimDetector/simDetectorApp/op' (728K)...
Removing directory '/opt/epics/modules/areaDetector/ADSimDetector/docs' (856K)...
Removing directory '/opt/epics/modules/areaDetector/ADSimDetector/dbd' (8.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADPilatus' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADPerkinElmer' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADPointGrey' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADCore/ci' (16K)...
Removing directory '/opt/epics/modules/areaDetector/ADCore/configure' (36K)...
Removing directory '/opt/epics/modules/areaDetector/ADCore/documentation' (6.8M)...
Removing directory '/opt/epics/modules/areaDetector/ADCore/XML_schema' (52K)...
Removing directory '/opt/epics/modules/areaDetector/ADCore/cfg' (16K)...
Removing directory '/opt/epics/modules/areaDetector/ADCore/include' (1.3M)...
Removing directory '/opt/epics/modules/areaDetector/ADCore/ADApp/ntndArrayConverterSrc' (40K)...
Removing directory '/opt/epics/modules/areaDetector/ADCore/ADApp/op' (16M)...
Removing directory '/opt/epics/modules/areaDetector/ADCore/ADApp/pluginTests' (308K)...
Removing directory '/opt/epics/modules/areaDetector/ADCore/ADApp/pluginSrc' (2.4M)...
Removing directory '/opt/epics/modules/areaDetector/ADCore/ADApp/ADSrc' (256K)...
Removing directory '/opt/epics/modules/areaDetector/ADCore/docs' (5.5M)...
Removing directory '/opt/epics/modules/areaDetector/ADCore/dbd' (44K)...
Removing directory '/opt/epics/modules/areaDetector/ADCameraLink' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADRoper' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/aravisGigE' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADAndor3' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADAndor' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADEiger' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADPluginEdge' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADFastCCD' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADCompVision' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADADSC' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADVimba' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADMythen' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADPvCam' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADmar345' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/firewireDCAM' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADURL' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADViewers' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/specsAnalyser' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADmarCCD' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADDexela' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADFireWireWin' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/docs' (412K)...
Removing directory '/opt/epics/modules/areaDetector/ADMerlin' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADQImaging' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/NDDriverStdArrays' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADLightField' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ffmpegViewer' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADMMPAD' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ffmpegServer' (4.0K)...
Removing directory '/opt/epics/modules/areaDetector/ADLambda' (4.0K)...
Pruning module '/opt/epics/modules/areaDetector/ADCore'...
Pruning module '/opt/epics/modules/areaDetector/ADSimDetector'...
Pruning module '/opt/epics/modules/areaDetector/ADSupport'...
```

</details>

base/lnls-prune-artifacts.sh
```diff
@@ -0,0 +1,113 @@
...
+set -Eeu
+
+# Filter out from the $1 list of paths any exact match of a parent directory
+# from any path in the $2 exclude list.
+#
+# Both list are treated as newline-separated strings, and must consist
+# exclusively of absolute paths.
+filter_out_paths() {
+    list="$1"
+    exclude_list=$(echo "$2" | grep -v '^$')
```
ericonr, line 12: I think if we accept a list with empty lines for second argument, we should try to do the same for the first one too.

ericonr (changes requested):
Considering the principle of each commit keeping the repository fully working, would it be possible to have "ioc: don't prune pre-selected directories." before "ioc: prune module directories from non-static builds."? Something like implement `get_defined_paths_to_keep`, then add it to relevant modules, then implement module pruning already using it? Would that make sense and be worth it?

I agree that the message for "base: don't prune HDF5 filter plugins from ADSupport." should mention the need to have the file in AreaDetector

Commit message nits:

> The set difference is works by

s/is//

> It is also susceptible to be skipped during module directory pruning if a target specifies it.

It can also be skipped

gustavosr8:
nit on 264776462a7bf02c531c3657a6e00d4fc06c2cc1 message:
'...to give users a immediate last resort...' -> '...to give users an immediate last resort...'

gustavosr8 (approved):

base/lnls-prune-artifacts.sh
henriquesimoes: Fair enough. I've also included a mention in the docstring.

henriquesimoes:
Thanks, @gustavosr8 and @ericonr. I've fixed all typos and implemented the minor suggestions.

> Considering the principle of each commit keeping the repository fully working, would it be possible to have "ioc: don't prune pre-selected directories." before "ioc: prune module directories from non-static builds."?

Conceptually, it makes sense. First, add the required tooling, then correctly prune things using them.

> Something like implement get_defined_paths_to_keep, then add it to relevant modules, then implement module pruning already using it? Would that make sense and be worth it?

As to whether it is worth it or not, I'm not sure. I'll give it a try and see if that's too much of a rework. At a first glance, it does not look like so, as "ioc: don't prune pre-selected directories." is pretty much self-contained and its usage is really simple. Refactoring messages is what will require a more careful look, I assume.

henriquesimoes:
Final diff is identical, as expected. I recommend using `git-range-diff(1)` for reviewing message changes ;)

Feel free to rebase and improve any of the messages at your will.

ericonr (approved):

PR #61, opened 2024-04-25
images: add OPCUA image.

ericonr:
Since the OPCUA IOC we use is built alongside the OPCUA module, we can simply export the IOC image ourselves.

---

This is still missing automatic generation of the image in CI. Creating the tagged image is easy, we just need to [depend](https://stackoverflow.com/questions/58457140/dependencies-between-workflows-on-github-actions) on the workflow which pushes the base image. The hard part is running it for PRs as well... This means it probably has to be in the same workflow and must deal with the recently built base image which will have its own name but should be called `X.Y.Z-dev`.

I'm not sure how far we need to go before merging and tagging a release so it can be used.

### DIFF
--- a/images/docker-compose-opcua.yml
+++ b/images/docker-compose-opcua.yml
@@ -1,0 +1,13 @@
+services:
+  ioc:
+    image: ghcr.io/cnpem/opcua-epics-ioc
+    build:
+      context: ./
+      dockerfile: ../Dockerfile
+      target: no-build
+      labels:
+        org.opencontainers.image.source: https://github.com/cnpem/epics-in-docker
+      args:
+        REPONAME: opcua
+        RUNDIR: /opt/epics/modules/opcua/iocBoot/iocUaDemoServer
+        RUNTIME_PACKAGES: libxml2 libssl1.1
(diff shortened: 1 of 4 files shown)


images/docker-compose-opcua.yml
```diff
@@ -0,0 +1,13 @@
...
+    image: ghcr.io/cnpem/opcua-epics-ioc
+    build:
+      context: ./
+      dockerfile: ../Dockerfile
+      target: no-build
+      labels:
+        org.opencontainers.image.source: https://github.com/cnpem/epics-in-docker
+      args:
+        REPONAME: opcua
+        RUNDIR: /opt/epics/modules/opcua
```
henriquesimoes, line 12: If we are going to use #59, we should be careful when defining these variables. By implementing the suggestion of dropping the `RUNDIR` as input to keep binaries, `REPONAME` should point to `epics/modules/opcua` instead. There is no such `/opt/opcua` directory in `no-build` targets as of now.

An alternative approach could be to symbolic link or hard-copy during the pruning stage of `no-build`.

henriquesimoes (commented):
No strong opinion regarding CI for these images for now. I'll think more about it.

henriquesimoes (changes requested):
Missing entry in CHANGES.md. It may be useful to specify such kind of image somewhere in the README.

henriquesimoes (approved):
Looks good.

PR #62, opened 2024-05-06
Support to uTCA IOC  

gustavosr8:


henriquesimoes (changes requested):
Please, follow the pattern for commit header we use, that is:

```
base: add iocStats module.
base: add ipmiComm module.
```

We use `base:` for commits affecting the base image and `ioc:` for commits affecting the IOC images. We can always `git log` the file and see the past messages touching it.

Moreover, for each commit, add a item in the `CHANGES.md` file under the Unreleased section following the current pattern.

henriquesimoes (changes requested):
We're almost there, I think.

Typo in the 9a7e6ca96959d178b8e36ca3c1e4a5fd18339e60: `s/The patches solves/The patch solves`.

base/docker-compose.yml
```diff
@@ -31,3 +31,4 @@ services:
         MOTOR_VERSION: ${MOTOR_VERSION}
         PMAC_VERSION: ${PMAC_VERSION}
         OPCUA_VERSION: ${OPCUA_VERSION}
+        IOCSTATS_VERSION: ${IOCSTATS_VERSION}
```
henriquesimoes, line 34: Noticed this now. But move this right below `ETHER_IP` to keep the order from the usage inside scripts. Same thing for ipmiComm (below iocStats).

henriquesimoes (commented):
Looks fine by me. Adding @ericonr as reviewer for a second look. You may wait for his review before resolving the following point I noticed only now.

base/install_modules.sh
```diff
@@ -65,3 +65,7 @@ EPICS_BASE
 install_from_github -i epics-modules ether_ip ETHER_IP $ETHER_IP_VERSION "
 EPICS_BASE
 "
+
+install_from_github  epics-modules iocStats IOCADMIN $IOCSTATS_VERSION "
+EPICS_BASE
+"
```
ericonr, line 71: The expected name for this module is `DEVIOCSTATS`. That's the library name, it won't work otherwise for other users.

`IOCADMIN` is only referred to in https://github.com/slac-epics-modules/ipmiComm/blob/3611042d5ad8394bacd073481c6b510fa2805199/configure/RELEASE_SPEAR#L22 , but that's not used for the build, as far as I can see, and it's SPEAR specific (they are even using the custom name from SLAC's repo: https://github.com/slac-epics/iocAdmin ), as far as I could see it's not actually used. Did you get any errors without it?

The commit message is wrong. There is no iocAdmin library, [only `devIocStats`](https://github.com/search?q=repo%3Aepics-modules%2FiocStats%20LIBRARY_IOC&type=code). I think the commit should ideally be simply addition of the module, and it shouldn't require any explanations as to why...

CHANGES.md
```diff
@@ -8,6 +8,8 @@
   https://github.com/cnpem/epics-in-docker/pull/57
 * base: add IOCStats module. by @gustavosr8 in
   https://github.com/cnpem/epics-in-docker/pull/62
+* base: add IPMIComm module. by @gustavosr8 in
+  https://github.com/cnpem/epics-in-docker/pull/62
```
ericonr, line 12: Nit: I'd rather have a single line in the change list, but that's not mandatory.

ericonr (changes requested):

henriquesimoes (approved):
It looks good to me. Thanks for working on this.

Just to be sure. Have you tested this latest version with https://github.com/lnls-dig/ipmi-mgr-epics-ioc?

base/install_modules.sh
```diff
@@ -69,3 +69,11 @@ EPICS_BASE
 install_from_github  epics-modules iocStats DEVIOCSTATS $IOCSTATS_VERSION "
 EPICS_BASE
 "
+
+download_from_github slac-epics-modules ipmiComm $IPMICOMM_VERSION
+patch -d ipmiComm -Np1 < ipmicomm.patch
+install_module ipmiComm IPMICOMM "
+EPICS_BASE
+ASYN
+DEVIOCSTATS
```
ericonr, line 78: As far as I can tell `DEVIOCSTATS` isn't needed by it either. Did you see it being required anywhere?

ericonr (commented):
The commit message has "patche" instead of "patch".

PR #63, opened 2024-05-10
images: add missing TAG to OPCUA image name.

ericonr:


henriquesimoes (approved):

henriquesimoes:
Just a question regarding the tag: should we concern about events?

```${{ github.event_name == 'push' && github.ref_name || github.head_ref }}```

ericonr:
For now it's only intended to run for tags; we can add that conditional when it can run properly for PRs as well.

PR #64, opened 2024-05-16
Pmac module version update, add motorPIGCS2 and MCA IOC

guirodrigueslima:


base/.env
```diff
@@ -22,6 +22,6 @@ NDSSCPIMEGA_VERSION=1.0.0
 LIBSSCPIMEGA_VERSION=fb8acf533a7c01b5266bf32d60d1a5f923e19523
 
 MOTOR_VERSION=R7-3-1
-PMAC_VERSION=2-6-1
+PMAC_VERSION=2-6-4b3
```
ericonr: As far as I can tell, these look like pre-releases. Do we need any of the included changes?

CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* base: Pmac module version update. by @guirodrigueslima
```
ericonr: This should be under a "Bug fixes" or "New features" subheader, and is missing a link to the PR.

CHANGES.md
```diff
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* base: Pmac module version update. by @guirodrigueslima
+* base: Pmac module version update and Add motorPIGCS2 IOC. by @guirodrigueslima
```
ericonr: "and add motorPIGCS2 IOC and image"

images/docker-compose-motorpigcs2.yml
```diff
@@ -0,0 +1,13 @@
+services:
+  ioc:
+    image: ghcr.io/cnpem/motorpigcs2-epics-ioc:$TAG
+    build:
+      context: ./
+      dockerfile: ../Dockerfile
+      target: no-build
```
ericonr, line 7: This IOC should support static linking, but we are using dynamic linking for it unnecessarily, in a way...

I think we could have an option to build these built-in IOCs statically.

images/docker-compose-motorpigcs2.yml
```diff
@@ -0,0 +1,13 @@
...
+      labels:
+        org.opencontainers.image.source: https://github.com/cnpem/epics-in-docker
+      args:
+        REPONAME: motorpigcs2
+        RUNDIR: /opt/epics/modules/motor/modules/motorPIGCS2/iocs/pigcs2IOC/iocBoot/iocPIGCS2
+        RUNTIME_PACKAGES:
```
ericonr, line 13: You can simply omit the empty variable.

base/.env
```diff
@@ -22,6 +22,7 @@ NDSSCPIMEGA_VERSION=1.0.0
 LIBSSCPIMEGA_VERSION=fb8acf533a7c01b5266bf32d60d1a5f923e19523
 
 MOTOR_VERSION=R7-3-1
+PIGCS2_VERSION=60af8bdb17c1717e4545d8170f820e358ce31458
```
ericonr, line 27: Do you think we could ask someone to add a tag to the module? It's okay to have the commit, but a tag is usually better :)

.github/workflows/included-images.yml
```diff
@@ -29,5 +29,5 @@ jobs:
         uses: docker/bake-action@v4
         with:
           workdir: images/
-          files: docker-compose-opcua.yml
+          files: docker-compose-opcua.yml docker-compose-motorpigcs2.yml
```
ericonr: I think we should use linebreaks in this entry, otherwise it will get too full.

base/.env
```diff
@@ -16,6 +16,8 @@ IPAC_VERSION=2.16
 CAPUTLOG_VERSION=R4.0
 RETOOLS_VERSION=b7abe82533cdbd7ddbc2dc845c95a2c51b7f2db9
 ETHER_IP_VERSION=ether_ip-3-3
+SCALER_VERSION=4.1
```
ericonr, line 19: Adding scaler should be a separate commit.

base/docker-compose.yml
```diff
@@ -29,5 +29,8 @@ services:
         NDSSCPIMEGA_VERSION: ${NDSSCPIMEGA_VERSION}
         LIBSSCPIMEGA_VERSION: ${LIBSSCPIMEGA_VERSION}
         MOTOR_VERSION: ${MOTOR_VERSION}
+        PIGCS2_VERSION: ${PIGCS2_VERSION}
```
ericonr, line 32: This is in the wrong commit.

base/install-functions.sh
ericonr, line 1: Changes to this file should be a separate commit.

base/install_modules.sh
```diff
@@ -65,3 +65,20 @@ EPICS_BASE
 install_from_github -i epics-modules ether_ip ETHER_IP $ETHER_IP_VERSION "
 EPICS_BASE
 "
+
+install_from_github -i epics-modules scaler SCALER $SCALER_VERSION "
+EPICS_BASE
+ASYN
+"
```
ericonr, line 72: Is there a scaler IOC?

images/docker-compose-mca.yml
```diff
@@ -0,0 +1,13 @@
+services:
+  ioc:
+    image: ghcr.io/cnpem/mca-epics-ioc:$TAG
+    build:
+      context: ./
+      dockerfile: ../Dockerfile
+      target: no-build
```
ericonr, line 7: Same thing here, I think we should support static linking the IOCs... (though they would probably remain `no-build` here)

It doesn't have to be changed right now, though.

ericonr (changes requested):
How soon do you need these?

It would be nice to have module pruning added *before* we merge so many new images.

base/.env
guirodrigueslima: There was no need for changes, just updating the version with corrections to the module.

images/docker-compose-motorpigcs2.yml
guirodrigueslima: It was not possible to update to "static-link", a failure is generated in the Dockerfile when copying the RELEASE file.

images/docker-compose-mca.yml
guirodrigueslima: It was not possible to update to "static-link", a failure is generated in the Dockerfile when copying the RELEASE file.

base/.env
guirodrigueslima: I will talk to the person responsible about the changes.

base/.env
ericonr: I meant, do we need any of the changes between 2-6-1 and 2-6-4b3? Could we use 2-6-3, or wait for the release of 2-6-4 proper?

images/docker-compose-motorpigcs2.yml
ericonr: Yeah, I think it won't be possible for now, we need a feature to support it.

base/docker-compose.yml
```diff
@@ -32,3 +32,5 @@ services:
         PIGCS2_VERSION: ${PIGCS2_VERSION}
         PMAC_VERSION: ${PMAC_VERSION}
         OPCUA_VERSION: ${OPCUA_VERSION}
+        SCALER_VERSION: ${SCALER_VERSION}
```
henriquesimoes, line 35: This is in the wrong commit. It should be in ef8d83cd30f752ea8accffe7adfae99f8efe32bd.

base/musl/Dockerfile
```diff
@@ -22,7 +22,10 @@ RUN apk add --no-cache \
     meson \
     re2c \
     readline-dev \
-    readline-static
+    readline-static \
+    libnet-dev \
+    libpcap-dev \
+    libusb-dev
```
henriquesimoes, line 28: Put these along with other libraries, alphabetically sorted.

base/musl/docker-compose.yml
```diff
@@ -25,3 +25,5 @@ services:
         CAPUTLOG_VERSION: ${CAPUTLOG_VERSION}
         RETOOLS_VERSION: ${RETOOLS_VERSION}
         ETHER_IP_VERSION: ${ETHER_IP_VERSION}
+        SCALER_VERSION: ${SCALER_VERSION}
```
henriquesimoes, line 28: Again, it should be in ef8d83cd30f752ea8accffe7adfae99f8efe32bd.

CHANGES.md
```diff
@@ -4,8 +4,8 @@
 
 ### New features
 
-* base: Pmac module version update, add motorPIGCS2 IOC and
-  add scaler IOC, and image.
+* base: Pmac module version update, add motorPIGCS2 IOC,
+  scaler IOC and MCA IOC, and images.
   by @guirodrigueslima in https://github.com/cnpem/epics-in-docker/pull/64
```
henriquesimoes, line 9: Maybe splitting this into multiple items makes it easier to read. What do you think?

base/.env
henriquesimoes: > Do we need any of the included changes?

Yes, we need https://github.com/dls-controls/pmac/commit/7ab9f2413d60ce77c1ba4c32c256253e0535ef7e, which prevents a segfault. @guirodrigueslima told me the problem possibly happens when disconnection a occurs (even though we haven't checked this in details).

> wait for the release of 2-6-4 proper?

@guirodrigueslima can answer this better than me, but I understand he wants to deploy this ASAP.

henriquesimoes (changes requested):

base/.env
guirodrigueslima: There are two problems with the current version (2-6-1), the first is a "segmentation fault", when the coordinate system axes object does not exist, I don't know how to reproduce this problem, it was observed by Diamond Light Source. The other problem is related to the loss of connection with delta tau, when connecting again the offsets receive null values.

ericonr: In that case, I think specifically the Pmac update would fit best under "Bug fixes".

And mention in the commit message that we are using a pre-release because we need the specific fixes and there hasn't been a release yet.

ericonr (changes requested):

.github/workflows/included-images.yml
```diff
@@ -30,4 +30,5 @@ jobs:
         with:
           workdir: images/
           files: docker-compose-opcua.yml
+                 docker-compose-motorpigcs2.yml
```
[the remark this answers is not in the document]
guirodrigueslima: If "\" or "- docker..." is used, the .yml file shows a syntax error on the corresponding lines. CI is not building this code ?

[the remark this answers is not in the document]
guirodrigueslima: Using "\" or "-" didn't work. I checked with @henriquesimoes, my VS Code is linked with Github Workflow, this syntax is correct. Another alternative is to use "|" in the first line.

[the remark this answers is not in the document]
henriquesimoes: `files` key seems not to be a YAML array proper, but rather [a newline-delimited string][1].

[1]: https://github.com/docker/bake-action/tree/v4?tab=readme-ov-file#inputs

base/install-functions.sh
ericonr: ping here, this is unresolved

base/musl/docker-compose.yml
```diff
@@ -28,3 +28,5 @@ services:
         IOCSTATS_VERSION: ${IOCSTATS_VERSION}
         IPMICOMM_VERSION: ${IPMICOMM_VERSION}
         SCALER_VERSION : ${SCALER_VERSION}
+        MCA_VERSION: ${MCA_VERSION}
+
```
ericonr, line 32: Extra newline.

ericonr (changes requested):

base/install-functions.sh
guirodrigueslima: Sorry, I skipped

CHANGES.md
```diff
@@ -28,6 +28,13 @@
 
 OPCUA container image build with GitHub Actions has been fixed.
 
+### New features
+
+* base: Add motorPIGCS2 IOC and module, and image. by @guirodrigueslima in
+  https://github.com/cnpem/epics-in-docker/pull/64
+* base: Add scaler module. by @guirodrigueslima in 
+  https://github.com/cnpem/epics-in-docker/pull/64
```
henriquesimoes, line 36: It looks like you've squashed the "base: add scaler module." commit in the rebase process.

base/install_motor.sh
```diff
@@ -10,11 +10,17 @@ git clone --depth 1 --branch ${MOTOR_VERSION} \
 cd motor/modules
 
 git submodule update --init --depth 1 -j ${JOBS} \
-    motorMotorSim
+    motorMotorSim \
+
```
henriquesimoes, line 14: Drop this extra `\`.

images/docker-compose-motorpigcs2.yml
```diff
@@ -0,0 +1,12 @@
+services:
+  ioc:
+    image: ghcr.io/cnpem/motorpigcs2-epics-ioc:$TAG
```
henriquesimoes, line 3: Name this `ghcr.io/cnpem/motor-pigcs2-epics-ioc:$TAG`. This will make motor IOCs follow a similar convention that we use with areaDetector IOCs, whose names are `ad-<module>-epics-ioc`.

henriquesimoes (changes requested):

CHANGES.md
henriquesimoes: I've noticed now that both changes are also in `v0.8.1` instead of `Unreleased`.

Put them along with the pmac entry in the *end* of the New features section of the Unreleased version. This makes the changelog ordered by changes that were introduced in that specific release.

henriquesimoes:
I've complemented the commit messages, and reworded somethings to be consistent with our previous commits.

I've also [recreated the scaler module commit](https://github.com/cnpem/epics-in-docker/pull/64/commits/bae937408cf03c62312d2635464a997a3baa317e), which was squashed by mistake during the latest rebase.

gustavosr8:
Typo in 913c7786ffc53dfb94e5c65d59afec8c2269cf8f commit message

> Some IOCs ~need~ needs their own path location variable in the RELEASE file
during build. To avoid custom installation procedures for them, add
"dependency_name" to the EPICS_RELEASE_FILE before ~ovewriting~ overwriting the module
RELEASE file during `install_module`.

gustavosr8:
I don't think I get the message in 4834166d3ebb883da7eb93fe9df72f34c8aff4fe, but I think it may be re-written like this:

> Build motorPIGCS2 module from epics-modules/motor. Using a specific
commit is necessary ~use the latest changes not yet released~ because 
the latest changes have not been yet released, but that have been running 
in deploy at Carnauba for almost an year.

base/Dockerfile
```diff
@@ -12,6 +12,8 @@ RUN apt update -y && \
         git \
         libaravis-dev \
         libevent-dev \
+        libnet-dev \
+        libpcap-dev \
```
gustavosr8, line 16: It may be better if you explain in the longer commit message why do you added this packages. (same in alpine Dockerfile)

henriquesimoes: In general, such dependencies are directly related to the libraries used in the source code. So, I usually don't bother to make this explicit. In this case, only a single device type (which we do not currently have) requires such libraries, so I've added a mention to.

Thanks for the question, nevertheless. I noticed we did not install the runtime packages for these two. Attempting to run with a Canberra device would not work.

gustavosr8 (approved):

henriquesimoes (approved):

PR #66, opened 2024-06-19
Pydevsup support

gustavosr8:


base/install_pydevsup.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -ex
+
+. /opt/epics/install-functions.sh
+
+download_from_github mdavidsaver pyDevSup $PYDEVSUP_VERSION
+echo PYTHON=python3 >> pyDevSup/configure/CONFIG_SITE
+install_module pyDevSup PYDEVSUP "
+EPICS_BASE
+"
```
ericonr, line 11: Is this a separate file because it's not working with Alpine, so couldn't be included in `install_modules.sh`?

Dockerfile
ericonr, line 1: The new arguments should be documented in the README.

ericonr (commented):
The commit adding the python dependencies should be the same as the one adding the module, IMO.

Dockerfile
```diff
@@ -44,12 +50,18 @@ FROM build-image AS build-stage
...
 RUN if [ -n "$BUILD_PACKAGES" ]; then \
         apt update && \
         apt install -y --no-install-recommends $BUILD_PACKAGES; \
     fi
 RUN lnls-get-n-unpack -r $BUILD_TAR_PACKAGES
+RUN if [ -n "$BUILD_PIP_PACKAGES" ]; then \
+        apt update && \
+        apt install -y --no-install-recommends pip && \
+        pip install $BUILD_PIP_PACKAGES; \
+    fi
```
ericonr, line 64: Do we have a case where we already know this is necessary? It's adding more code and args which we won't even be testing...

README.md
```diff
@@ -65,8 +65,9 @@ stages even when they are not needed.
 Additional build and runtime packages to be installed can be listed in `args`,
 under the `BUILD_PACKAGES` and `RUNTIME_PACKAGES` keys, respectively. It is not
 necessary to quote them - e.g. `BUILD_PACKAGES: python3 python3-requests`.
-Packages essential to all (or most) IOCs should be added to [this repository's
-`Dockerfile`](./Dockerfile).
+Packages that strictly need to be installed via pip can be listed under the
+`BUILD_PIP_PACKAGES` and `RUNTIME_PIP_PACKAGES` keys. Packages essential to all
+(or most) IOCs should be added to [this repository's `Dockerfile`](./Dockerfile).
```
ericonr, line 70: "via `pip`"

Dockerfile
```diff
@@ -24,6 +25,13 @@ COPY --from=build-image /usr/local/lib /usr/local/lib
 COPY --from=build-image /usr/local/bin/lnls-get-n-unpack /usr/local/bin/lnls-get-n-unpack
 RUN lnls-get-n-unpack -r $RUNTIME_TAR_PACKAGES && \
     ldconfig
+RUN if [ -n "$RUNTIME_PIP_PACKAGES" ]; then \
+        apt update && \
+        apt install -y --no-install-recommends pip && \
+        pip install $RUNTIME_PIP_PACKAGES; \
+    apt clean && \
+    rm -rf /var/lib/apt/lists/*
+    fi
```
ericonr, line 34: What do you think about adding a line like `$([ -n "$RUNTIME_PIP_PACKAGES" ] && echo pip)` to the package list above?

Dockerfile
```diff
@@ -44,6 +52,7 @@ FROM build-image AS build-stage
 ARG REPONAME
 ARG BUILD_PACKAGES
 ARG BUILD_TAR_PACKAGES
+ARG BUILD_PIP_PACKAGES
```
ericonr, line 55: Leftover argument.

README.md
```diff
@@ -65,8 +65,9 @@ stages even when they are not needed.
 Additional build and runtime packages to be installed can be listed in `args`,
 under the `BUILD_PACKAGES` and `RUNTIME_PACKAGES` keys, respectively. It is not
 necessary to quote them - e.g. `BUILD_PACKAGES: python3 python3-requests`.
-Packages essential to all (or most) IOCs should be added to [this repository's
-`Dockerfile`](./Dockerfile).
+Packages that strictly need to be installed via `pip` can be listed under the
+`BUILD_PIP_PACKAGES` and `RUNTIME_PIP_PACKAGES` keys. Packages essential to all
```
ericonr, line 69: Leftover docs for build.

ericonr (commented):
I think the last commit could also update the changelog entry to mention support for installing pip packages.

Dockerfile
ericonr: To explain this, I'd rather not repeat the commands for apt.

Dockerfile
```diff
@@ -24,6 +26,11 @@ COPY --from=build-image /usr/local/lib /usr/local/lib
 COPY --from=build-image /usr/local/bin/lnls-get-n-unpack /usr/local/bin/lnls-get-n-unpack
 RUN lnls-get-n-unpack -r $RUNTIME_TAR_PACKAGES && \
     ldconfig
+RUN if [ -n "$RUNTIME_PIP_PACKAGES" ]; then \
+        pip install $RUNTIME_PIP_PACKAGES; \
+    apt clean && \
+    rm -rf /var/lib/apt/lists/*
```
ericonr, line 32: You can now clean up these apt commands

README.md
```diff
@@ -65,8 +65,10 @@ stages even when they are not needed.
 Additional build and runtime packages to be installed can be listed in `args`,
 under the `BUILD_PACKAGES` and `RUNTIME_PACKAGES` keys, respectively. It is not
 necessary to quote them - e.g. `BUILD_PACKAGES: python3 python3-requests`.
-Packages essential to all (or most) IOCs should be added to [this repository's
-`Dockerfile`](./Dockerfile).
+Packages that strictly need to be installed via `pip` can be listed under the
+`RUNTIME_PIP_PACKAGES` key and `pip` should be listed under `RUNTIME_PACKAGES`.
+Packages essential to all (or most) IOCs should be added to
```
ericonr, line 70: We don't need `pip` in `RUNTIME_PACKAGES` now, it's added automatically.

ericonr (changes requested):
I suggested some changes. Fixing those, I think we can merge :)

ericonr (approved):

PR #67, opened 2024-07-24
base: add SNMP module

guirodrigueslima:
Added the SNMP module, developed by SLAC (National Accelerator Laboratory). This module will be used to develop the IOC for Aruba Switches to control PoE ports.


henriquesimoes (approved):
As we have discussed in person, there are two repositories for SNMP at [slac-epics](https://github.com/slac-epics):

- [snmp](https://github.com/slac-epics/snmp)
- [epics-snmp](https://github.com/slac-epics/epics-snmp)

Both of them seem to implement the support for SNMP. Taking a brief look at them, the latter has been changed on GitHub more recently (6 months versus 3 years ago), but [supports less record types](https://github.com/slac-epics/epics-snmp/blob/12a9cfa7b5b07bee17ae8ba7a74b460dd72811e6/snmpApp/src/devSnmp.dbd) than [the former](https://github.com/slac-epics/snmp/blob/134fbfd91ac83974730093330781cd1217017382/devSnmpApp/src/devSnmp.dbd). `slac-epics/epics-snmp` also has more custom configurations in its build, which would require us to override them.

Considering that:

- we have tested with slac-epics/snmp and it has worked just fine for the simple IOC we need;
- the IOC is simpler with the mbbo records we would lose with epics-snmp;
- the build required no custom handling from our part

IMHO, we can stick with it, at least for now. If it turns out to be deprecated by SLAC or the other becomes more appealing, we can reconsider this choice.

henriquesimoes:
I've complemented the commit message with the relevant points regarding the choice of using slac-epics/snmp. Can you double check if everything is okay with it?

henriquesimoes:
I've asked in [tech-talk about the other modules][1] and for recommendations to see if this is indeed the best choice.

Let's see the community point of view of this.

[1]: https://epics.anl.gov/tech-talk/2024/msg00889.php

henriquesimoes:
@guirodrigueslima, can you take a look in the changes I've made?

They are mostly documentation changes, apart from some [changes to make `unzip` behave as `tar`](https://github.com/cnpem/epics-in-docker/compare/8323d7911886c5ee1b05aa010b0540754c3ed445..32aa2c9573978e1cfb8c6bf7114c908caf5740a0). I had to test if they really worked, so I edited the commit directly instead of asking for changes during review.

I think we are good to merge after that.

gustavosr8 (approved):

base/lnls-get-n-unpack.sh
```diff
@@ -19,6 +19,14 @@ for url; do
...
     wget -P $download_dir -o /tmp/wget.log "$url" || (cat /tmp/wget.log && false)
-    tar --no-same-owner -xf $download_dir/* -C $dest
+
+    filename=$(basename $download_dir/*)
+
+    if [[ ${filename,,} == *".zip" ]]; then
+        unzip -qo $download_dir/$filename -d $dest
+    else
+        tar --no-same-owner -xf $download_dir/$filename -C $dest
+    fi
```
ericonr, line 29: I think if we end up having to support more formats, we should switch to bsdtar+libarchive directly, instead of having to handle different tools in the script. Since busybox already provides unzip, I think it's reasonable to depend on it.

PR #69, opened 2024-07-29
Ensure single UID is used inside containers

henriquesimoes:
APT is one of the applications that assume multiple effective UIDs exist, so that it can sandbox itself. When running with `ignore_chown_errors` option enabled (see [podman(1)](https://manpages.debian.org/bookworm/podman/podman.1.en.html#Rootless_mode)), this will lead to `setgroups`, `seteuid` and `setguid` syscalls to fail, breaking most of `apt` commands.

Ensure that a single UID is used inside the containers, both at base build image (without `/etc/set{u,g}id` settings) and IOC runtime, especially allowing one to properly use apt.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,8 @@ ARG RUNTIME_PACKAGES
 ARG RUNTIME_TAR_PACKAGES
 ARG RUNTIME_PIP_PACKAGES
 
+COPY --from=build-image /etc/apt/apt.conf.d/90-disable-sandbox.conf /etc/apt/apt.conf.d/90-disable-sandbox.conf
+
 RUN apt update -y && \
     apt install -y --no-install-recommends \
         libreadline8 \
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Bug fixes
+
+* ioc: disable APT sandbox. by @henriquesimoes in
+  https://github.com/cnpem/epics-in-docker/69
+  * This allows to use APT in containers deployed in systems without subuid and
+    subgid.
+
 ### New features
 
 * base: update ipmiComm patch to install general templates. by
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -6,6 +6,10 @@ ARG JOBS
 
 ENV DEBIAN_FRONTEND noninteractive
 
+# Disable APT sandbox so that single UID restriction is satisfied in systems
+# without subuid and subgid configured.
+RUN echo 'APT::Sandbox::User "root";' > /etc/apt/apt.conf.d/90-disable-sandbox.conf
+
 RUN apt update -y && \
     apt install -y --no-install-recommends \
         build-essential \


base/lnls-single-uid.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+#
+# Configure container filesystem so that single UID restriction is satisfied,
+# making ignore_chown_errors=True storage configuration work properly.
+
+echo "APT::Sandbox::User \"root\";" > /etc/apt/apt.conf.d/90-disable-sandbox.conf
```
henriquesimoes, line 6: Maybe this is not enough, since some files maybe owned by `adm` by default. With the current implementation, APT will still complain about `term.log` file:

```
W: chown to root:adm of file /var/log/apt/term.log failed - OpenLog (22: Invalid argument)
```

ericonr: That's a warning though, right?

Dockerfile
```diff
@@ -10,6 +10,9 @@ ARG RUNTIME_PACKAGES
 ARG RUNTIME_TAR_PACKAGES
 ARG RUNTIME_PIP_PACKAGES
 
+COPY --from=build-image /usr/local/bin/lnls-single-uid /usr/local/bin/lnls-single-uid
+RUN lnls-single-uid
+
```
ericonr, line 15: I'm not sure I agree with this being a `COPY`. (including for the base image)

It's such a short command...

I'd be okay with this copying the rule from the base image, though.

base/lnls-single-uid.sh
ericonr, line 6: Use single quotes for the outer quotes, to avoid needing to escape the inner ones.

base/lnls-single-uid.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+#
+# Configure container filesystem so that single UID restriction is satisfied,
+# making ignore_chown_errors=True storage configuration work properly.
```
ericonr, line 4: IMO this is the sort of information that one can get directly from git history; it's clearly being done for a reason and someone who wishes to remove/understand it can git blame.

I don't know if I'm going too far on the frugal scale for docs, though, so if you'd rather keep we can leave it in.

Either way, `ignore_chown_errors=True` is not the correct configuration option. It's "systems without subuid and subgid configured"; `ignore_chown_errors=True` is part of what makes them work, but not what defines them.

base/lnls-single-uid.sh
henriquesimoes: Will do, thanks.

henriquesimoes: Yes, it is. I wonder if there is anything else we need to configure to properly have all its important features properly working. It seems that logging will be somewhat broken.

I haven't done any research on the topic yet.

henriquesimoes: Well, there's nothing we can do regarding the log file warning, as this [attempt to change group ownership is hard-coded][1]. But it is indeed harmless.

I could not see other files with different ownership when sub{u,g}id are defined. It seems [other sandbox options][2] do not need to be changed as well.

[1]: https://salsa.debian.org/apt-team/apt/-/blob/8a8fdea4c71fbddfda9f7d6a1d9ec5da68870e12/apt-pkg/deb/dpkgpm.cc#L1022
[2]: https://salsa.debian.org/apt-team/apt/-/blob/8a8fdea4c71fbddfda9f7d6a1d9ec5da68870e12/doc/examples/configure-index#L669-L685

ericonr (approved):

PR #72, opened 2024-08-20
ioc: install busybox symbolic links.

henriquesimoes:
Make all busybox goodies available in the PATH without having to worry about them being from busybox. This is done automatically for Alpine, but not for Debian.

This removes the need to install wget, since it is already provided by busybox. However, its version packaged by Debian does not implement `-o` flag, requiring us to use bash redirection instead.

Fixes: ffdbecc1 (ioc: initial commit.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,10 +16,9 @@ RUN apt update -y && \
         busybox \
         netcat-openbsd \
         procserv \
-        wget \
         $([ -n "$RUNTIME_PIP_PACKAGES" ] && echo pip) \
         $RUNTIME_PACKAGES && \
-    ln -s /bin/busybox /usr/bin/unzip && \
+    busybox --install && \
     apt clean && \
     rm -rf /var/lib/apt/lists/*
 
--- a/README.md
+++ b/README.md
@@ -72,6 +72,11 @@ Packages that strictly need to be installed via `pip` can be listed under the
 Packages essential to all (or most) IOCs should be added to
 [this repository's `Dockerfile`](./Dockerfile).
 
+Extra files can also be downloaded and installed by listing their TAR or ZIP
+archive URLs under `BUILD_TAR_PACKAGES` or `RUNTIME_TAR_PACKAGES`. To use HTTPS
+(or any other TLS-based protocol), `ca-certificates` must also be added to
+`RUNTIME_PACKAGES`.
+
 The template above assumes the containers will be uploaded to the GitHub
 registry.
 
--- a/base/lnls-get-n-unpack.sh
+++ b/base/lnls-get-n-unpack.sh
@@ -18,7 +18,7 @@ for url; do
     download_dir=$(mktemp -d)
 
     echo Downloading "$url"...
-    wget -P $download_dir -o /tmp/wget.log "$url" || (cat /tmp/wget.log && false)
+    wget -P $download_dir "$url" &> /tmp/wget.log || (cat /tmp/wget.log && false)
 
     filename=$(basename $download_dir/*)
 


Dockerfile
```diff
@@ -16,10 +16,9 @@ RUN apt update -y && \
         busybox \
         netcat-openbsd \
         procserv \
-        wget \
         $([ -n "$RUNTIME_PIP_PACKAGES" ] && echo pip) \
         $RUNTIME_PACKAGES && \
-    ln -s /bin/busybox /usr/bin/unzip && \
+    busybox --install && \
```
ericonr, line 21: Does busybox wget support HTTPS?

ericonr (commented):
I'm not sure I agree this is a fix. Why do you think it is?

Dockerfile
henriquesimoes: On local tests I've done, it seems to support TLS, yes. [On their documentation page][1], there is nothing about this, though.

[1]: https://www.busybox.net/downloads/BusyBox.html

henriquesimoes:
> I'm not sure I agree this is a fix. Why do you think it is?

Because I bet no one would use it as `busybox <command>`. Most users would expect the command to be available in the PATH, and install it from the package registry when the they noticed the command did not exist. If I had noticed this is not the case earlier, I'd have proposed to have symlinks installed from the start.

But I understand that is mostly an extension of the features we provided in previous images, and thus not exactly a bug fix indeed. If you still think it better fits the "new features" section, let me know.

henriquesimoes:
> I think the main commit being fixed is the one that added the possibility of fetching stuff during build. Having done this for the initial busybox commit would have been helpful, but not critical. What do you think?

I understand you point, but I also meant users `podman exec`ing inside containers and attempting to debug something. In such cases, for instance, they would not need to install `iputils-ping`, but I think they would, because ping would not be in PATH. That's why I referenced to the initial commit instead of 8f02abc2047d5087ce69ed032b4e42006ec56605 .

> And I don't think it makes sense to mention alpine here, since we only use debian for the IOC images.

I agree on this one. Referencing it made sense for me at start because I was troubled about why we did not need to install it in alpine as we did for `lnls-debian-11-epics-7`.

Dockerfile
henriquesimoes: Users should be aware that they need to add `ca-certificates` to `RUNTIME_PACKAGES`, though. Debian image currently does not add it by default; there is an issue open for that: https://github.com/debuerreotype/docker-debian-artifacts/issues/15.

Maybe that would be something to document, but I'm not sure if it pays off.

README.md
```diff
@@ -68,9 +68,12 @@ Additional build and runtime packages to be installed can be listed in `args`,
...
 Packages that strictly need to be installed via `pip` can be listed under the
-`RUNTIME_PIP_PACKAGES` key.
-Packages essential to all (or most) IOCs should be added to
-[this repository's `Dockerfile`](./Dockerfile).
+`RUNTIME_PIP_PACKAGES` key. Extra files can be downloaded and installed to the
+container by listing their TAR or ZIP archive URLs under `BUILD_TAR_PACKAGES`
+or `RUNTIME_TAR_PACKAGES`. To use HTTPS (or any other TLS-based protocol),
+`ca-certificates` must also be added in `RUNTIME_PACKAGES`. Packages essential
+to all (or most) IOCs should be added to [this repository's
+`Dockerfile`](./Dockerfile).
```
ericonr, line 76: I would split the new content into its own paragraph. And keep the "Packages essential..." part where it used to be.

ericonr (changes requested):

README.md
henriquesimoes: > And keep the "Packages essential..." part where it used to be.

Do you mean right after pip install of things?

PR #73, opened 2024-08-26
base: update ipmicomm.patch to install general templates.

gustavosr8:


base/ipmicomm.patch
```diff
@@ -1,5 +1,24 @@
 From: https://github.com/slac-epics-modules/ipmiComm/pull/26/
+From: https://github.com/slac-epics-modules/ipmiComm/pull/27/
```
henriquesimoes, line 2: I think you wanted to reference https://github.com/slac-epics-modules/ipmiComm/pull/29 instead.

henriquesimoes (changes requested):

henriquesimoes (commented):
I'd take the opportunity and rename it to `backport-ipmicomm.patch`, since all those have been merged to the pre-release branch.

It seems sensible to add an entry in the change log as well.

ericonr (approved):
Approved after following @henriquesimoes 's suggestions.

henriquesimoes (approved):

PR #75, opened 2024-08-28
Supporting statically linked IOCs

ericonr:


base/musl/Dockerfile
```diff
@@ -45,7 +46,7 @@ COPY install-functions.sh .
 COPY backport-epics-base-musl.patch .
 COPY epics-base-static-linking.patch .
 COPY install_epics.sh .
-RUN ./install_epics.sh
+RUN COMMANDLINE_LIBRARY=READLINE_NCURSES ./install_epics.sh
```
henriquesimoes, line 49: So it turns out that we rely on [`__has_include` compiler feature to properly link to readline][1] instead of EPICS' implementation in the Debian image. This is not an option here though only because we need to also specify its dependency. That's what I understand about the problem being solved here.

Since older versions of libreadline required also linking directly to ncurses (?), [EPICS has personalized options][2] for that too, including `READLINE_NCURSES`. Thus, it turns out it just a coincidence that we can explore this to make musl build work. Is that right?

If so, it may be better to make explicit in the commit message that we are taking advantage of a option defined for another purpose, as it may be removed from EPICS base if someone decides to drop support for older Linux distros eventually.

[1]: https://github.com/epics-base/epics-base/blob/052a0c7e19580c18077a6fc5cf9f67ce75e3082b/documentation/RELEASE_NOTES.md#automatic-commandline_library-w-newer-compilers
[2]: https://github.com/epics-base/epics-base/blob/052a0c7e19580c18077a6fc5cf9f67ce75e3082b/configure/os/CONFIG.Common.linuxCommon#L38-L43

ericonr: You understood it correctly. I will update the commit to make what's going on clearer.

ericonr: @henriquesimoes lemme know what you think.

henriquesimoes: Seems nice and clear now. Thanks!

base/build-static-ioc.sh
```diff
@@ -0,0 +1,35 @@
...
+
+make_skip distclean
+make
+make_skip clean
+make -C iocBoot
+
+version=$(git describe)
+
+echo $version > $IOC-version
+echo ".git/" > /tmp/globs
```
ericonr, line 30: The necessary globs for building afc-epics-ioc correctly are still going to be determined.

gustavosr8 (commented):
Isn't better to squash this commit with the previous, keeping the changes tracking with the change itself?

base/build-static-ioc.sh
```diff
@@ -0,0 +1,35 @@
...
+    make "$@"
+  fi
+}
+
+make_skip distclean
+make
+make_skip clean
+make -C iocBoot
+
+version=$(git describe --tags)
```
gustavosr8, line 27: NIT: this command add the number of additional commits and an abbreviated object name for the commit itself at the end of the tag. Maybe using the flag `--abbrev=0` keeps the tarball name more clean.

ericonr: In this case I do want the object name. If it's a tag, the name is clean, if it's  a development version, we have all the necessary information about the version.

ericonr:
> Isn't better to squash [this commit](https://github.com/cnpem/epics-in-docker/pull/75/commits/a8ea191dbb7eacdd326757c481b2dd79ff21596c) with the previous, keeping the changes tracking with the change itself?

I'm not sure. The change is a combination of all the other commits in the PR, IMO. Even if the last commit was the one to tie everything together :thinking:

base/build-static-ioc.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
+
+set -ex
+
+IOC=$1
```
henriquesimoes, line 5: `$1` may be undefined, and the shell will happily accept it until `tar` fails to find the directory. I know this argument is clearly documented in the README, but it should be clear to understand what's going on when it is forgotten.

Adding `nounset (-u)` option above could fix this.

base/build-static-ioc.sh
```diff
@@ -0,0 +1,35 @@
...
+
+cp /opt/epics/RELEASE configure/
+cat << EOF > configure/CONFIG_SITE.local
+STATIC_BUILD=YES
+FULL_STATIC_BUILD=YES
+USR_LDFLAGS = -static-pie
+EOF
```
henriquesimoes, line 12: I feel like overwriting this can be an issue for some IOCs. I'm okay with it, but it seems worth documenting it.

base/musl/Dockerfile
```diff
@@ -73,3 +73,5 @@ COPY backport-ipmicomm.patch .
 COPY caputlog-waveform-fix.patch .
 COPY install_modules.sh .
 RUN NEEDS_TIRPC=YES ./install_modules.sh
+
+COPY build-static-ioc.sh /usr/local/bin
```
henriquesimoes, line 77: Other similar scripts have been prefixed with `lnls-`, and had their `.sh` prefix dropped when installing. I'd rather keep this consistent.

README.md
```diff
@@ -151,3 +151,31 @@ default image with older kernels.
...
+services:
+  build-static-ioc:
+    image: ghcr.io/cnpem/lnls-alpine-3.18-epics-7:v0.10.0-dev
+    volumes:
+      - type: bind
+        source: ./
+        target: /opt/REPOSITORY
+    working_dir: /opt/REPOSITORY
+    command: build-static-ioc.sh REPOSITORY
+```
```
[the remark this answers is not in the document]
ericonr: This is a general fix, we don't do it for other blocks. Will add it.

base/build-static-ioc.sh
ericonr: I thought I had added `-u`, actually. Thanks for spotting it :)

base/build-static-ioc.sh
ericonr: `CONFIG_SITE.local` shouldn't be checked into git. What sort of IOCs do you think this could be an issue for?

README.md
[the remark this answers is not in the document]
henriquesimoes: Yeah, we kinda missed it for CI workflow boilerplate. Debian-based IOC template is fine, though.

base/build-static-ioc.sh
henriquesimoes: Indeed. That doesn't mean people follow this, though. For instance, [`ether_ip` is a module that does not][1]... there are certainly [others around][2].

But my point is that someone can attempt to build any IOC, and suddenly all her well-crafted and not versioned configuration in`CONFIG_SITE.local` will be overwritten in her host machine, since it is volume mounted. She won't be able to `git restore` it, because it shouldn't even be tracked by Git in the first place.

Just to be clear, I don't see overwriting an issue for us, since we won't be using any local site configuration for both RFFE and AFC IOCs. But future users might have to use them for overwriting default settings they care about.

[1]: https://github.com/epics-modules/ether_ip/blob/6ca0e66ef72510afb51b4dfab987b0ef5d9b8d36/configure/CONFIG_SITE.local
[2]: https://github.com/search?q=path%3Aconfigure%2FCONFIG_SITE.local&type=Code&ref=advsearch&l=&l=

ericonr: This is somewhat related to an annoyance I have had with the RELEASE file being marked as changed in git...

Do you think it would be a good alternative to do the following?

1. backup `RELEASE` and `CONFIG_SITE.local`
2. add our own versions, and use them for build
3. restore them when the script exits and with a trap handler, in case one of the make commands errors out

ericonr: What do you think of this solution?

henriquesimoes: > What do you think of this solution?

I think we are almost there. It misses one thing: we should append to the end of existing `CONFIG_SITE.local`, and not create it entirely from scratch. After all, if someone maintains a local `CONFIG_SITE`, they have something they want to (or have to) customize in the build.

henriquesimoes: In a similar vein, I'd consider appending (`+=`) to `LD_FLAGS` instead of overwriting (`=`) it.

henriquesimoes:
> > Isn't better to squash [this commit](https://github.com/cnpem/epics-in-docker/pull/75/commits/a8ea191dbb7eacdd326757c481b2dd79ff21596c) with the previous, keeping the changes tracking with the change itself?
>
> I'm not sure. The change is a combination of all the other commits in the PR, IMO. Even if the last commit was the one to tie everything together 🤔

There is one thing that I don't like about letting them in the same commit: when reverting a commit, you need to be careful not to revert the changelog. Thus, this is necessarily partial revert. But you should probably also specify in the log that you did this, so not a big deal.

On the other hand, it is pretty handy for blaming changes.

For this case, I also understand this a whole PR change that is being described. So I'm okay with it as is.

base/build-static-ioc.sh
ericonr: > > What do you think of this solution?
> 
> I think we are almost there. It misses one thing: we should append to the end of existing `CONFIG_SITE.local`, and not create it entirely from scratch. After all, if someone maintains a local `CONFIG_SITE`, they have something they want to (or have to) customize in the build.

I don't know how much I agree with this :thinking: 

The opposite situation is a custom `CONFIG_SITE.local` for building outside of the container, and using that erroneously when you mean to build using just the options from the script.

> In a similar vein, I'd consider appending (+=) to LD_FLAGS instead of overwriting (=) it.

That's fair

henriquesimoes: > The opposite situation is a custom CONFIG_SITE.local for building outside of the container, and using that erroneously when you mean to build using just the options from the script.

Indeed. Not sure the best way to deal with this. In this case, the user has better control of the situation, I think. She can rename the file to preserve her custom out-of-container settings, and let it clean for the script. In the situation I had described, she could change the Git-tracked `CONFIG_SITE`, or write her own version of `lnls-build-static-ioc`.

None of them seems optimal in the end. I think the latter is messier as it involves changing a tracked file, or rewriting something non-trivial.

If you prefer, we can let this as is, and change this if someone gets annoyed by it in the future.

ericonr: Add `-include CONFIG_SITE.local.lnls-build-static-ioc` to the `CONFIG_SITE.local` I write?

gustavosr8: > > The opposite situation is a custom CONFIG_SITE.local for building outside of the container, and using that erroneously when you mean to build using just the options from the script.
> 
> Indeed. Not sure the best way to deal with this. In this case, the user has better control of the situation, I think. She can rename the file to preserve her custom out-of-container settings, and let it clean for the script. In the situation I had described, she could change the Git-tracked `CONFIG_SITE`, or write her own version of `lnls-build-static-ioc`.
> 
> None of them seems optimal in the end. I think the latter is messier as it involves changing a tracked file, or rewriting something non-trivial.
> 
> If you prefer, we can let this as is, and change this if someone gets annoyed by it in the future.

Regardless of the solution used, I think that all these decisions that the user has to make must be very well documented it README.md to prevent any erroneously use of this feature

henriquesimoes: > Add -include CONFIG_SITE.local.lnls-build-static-ioc to the CONFIG_SITE.local I write?

That seems a good option if we continue down this road.

> Regardless of the solution used, I think that all these decisions that the user has to make must be very well documented it README.md to prevent any erroneously use of this feature

Agreed.

ericonr:
@henriquesimoes I didn't copy the directory inside the container as we had discussed, for the reasons laid out in the commit message. Let me know what you think.

henriquesimoes (changes requested):

README.md
```diff
@@ -151,3 +151,35 @@ default image with older kernels.
...
+
+Fully static IOCs can be built using the `lnls-build-static-ioc` script
+provided by this image. One way to automate this is using a
+`docker-compose.yml` file and running `docker compose up`; the file should have
+the following contents:
+
+```yaml
+services:
+  build-static-ioc:
+    image: ghcr.io/cnpem/lnls-alpine-3.18-epics-7:v0.10.0-dev
```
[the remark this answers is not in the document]
ericonr: Fixed.

base/build-static-ioc.sh
henriquesimoes: Has this been resolved?

henriquesimoes (approved):
> I didn't copy the directory inside the container as we had discussed, for the reasons laid out in the commit message. Let me know what you think.

That's fine. It would be easier if EPICS supported out-of-tree compilation.

ericonr:
Assuming we merge #75 , this will require an update :)

henriquesimoes:
> Assuming we merge https://github.com/cnpem/epics-in-docker/pull/75 , this will require an update :)

You mean #77, right? We may do the other way around if you prefer.

ericonr:
> > Assuming we merge #75 , this will require an update :)
> 
> You mean #77, right? We may do the other way around if you prefer.

Oops, yeah. Thanks :)

I've already rebased #77 locally on top of this, so I'm pending towards merging this one first.

gustavosr8:
WDYT about adding `bash` to the following in e7eeebc?

https://github.com/cnpem/epics-in-docker/blob/b27a3b044e37fffafb1a1910651d5354bf9166d5/README.md?plain=1#L29-L31

henriquesimoes (approved):
Well, changes look right to me. Nice to know that the build system keeps everything inside `O.*`. It is good to merge for me.

> WDYT about adding bash to the following in https://github.com/cnpem/epics-in-docker/commit/e7eeebc4a7149f62ba9cdf2bb0db9e91e9152a23?

I have been disappointed by the syntax highlight for `bash` in GitHub. Not sure if `shell` behaves differently or if all of them are not properly rendered.

PR #76, opened 2024-09-03
base: update ipmiComm patch to handle `CODE_DESTINATION_UNAVAIL`.

gustavosr8:


henriquesimoes (changes requested):
This is not a backport yet. Please, place it in a different file instead. Also, refer to its upstream PR in the corresponding patch file.

For the commit message, I'd rather something shorter and more direct:

```
base: patch ipmiComm to handle CODE_DESTINATION_UNAVAIL.
```

CHANGES.md
```diff
@@ -15,6 +15,8 @@ the experience of using the images.
   https://github.com/cnpem/epics-in-docker/69
   * This allows to use APT in containers deployed in systems without subuid and
     subgid.
+* base: patch ipmiComm to handle `CODE_DESTINATION_UNAVAIL`. by
+  @gustavosr8 in https://github.com/cnpem/epics-in-docker/pull/76.
```
ericonr, line 19: The release notes are now in the wrong place.

ericonr (commented):
I think we should hold off on creating epics-in-docker releases due to ipmiComm until we have determined that everything we need is working.

ericonr (approved):
I'm not a fan of backticks in commit messages generally, especially titles.

Your choice if you want to change, though.

base/ipmicomm.patch
```diff
@@ -0,0 +1,14 @@
+From: https://github.com/slac-epics-modules/ipmiComm/pull/33/
+diff --git a/src/drvMch.c b/src/drvMch.c
+index 33e9b00..8a057d0 100644
+--- a/src/drvMch.c
++++ b/src/drvMch.c
```
henriquesimoes, line 5: I'm missing the explanation from your upstream patch. You can always download a `git format-patch`ed commit from GitHub by appending `.patch` to its URL, as such:

https://github.com/slac-epics-modules/ipmiComm/commit/d73ee36351ac3914ed780d432e2a03c382f2c349.patch

henriquesimoes (approved):

PR #84, opened 2024-10-31
base: update to Debian 12.

ericonr:
And update relevant package versions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-ARG DEBIAN_VERSION=11.9
+ARG DEBIAN_VERSION=12.8
 
 FROM ghcr.io/cnpem/lnls-debian-epics-7:v0.12.0-dev AS build-image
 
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -1,6 +1,6 @@
-ARG DEBIAN_VERSION=11.9
+ARG DEBIAN_VERSION=12.7
```
henriquesimoes, line 1: Do we really need to replicate this as a default value?

base/.env
```diff
@@ -1,4 +1,4 @@
-DEBIAN_VERSION=11.9
+DEBIAN_VERSION=12.7
```
henriquesimoes, line 1: 12.8 is already available ;)

henriquesimoes (changes requested):

Dockerfile
henriquesimoes: Never mind. I thought this was in the base image Dockerfile, but it is not. It will be harder to make both of them include the same definition indeed.

PR #85, opened 2024-10-31
EPICS 7.0.8.1 and opcua

ericonr:
This also made it necessary to update the OPCUA module, so we could keep using the prebuilt version.


base/install_opcua.sh
```diff
@@ -5,13 +5,19 @@ set -ex
...
+mv opcuaBinaryDist opcua-module
+install_module opcua-module OPCUA "
 EPICS_BASE
 "
 
+mv opcuaExampleIoc opcua
+install_module -i opcua OPCUA_IOC "
+EPICS_BASE
+OPCUA
+"
```
henriquesimoes, line 20: As discussed in person, mention that `opcua/` is kept for backward compatibility with existing iocsh scripts.

henriquesimoes (changes requested):

henriquesimoes (commented):
Nice. That seems to be in good shape now. I think it's worth adding an entry in the changelog for this, so we can know when a IOC can use newer EPICS 7 features. What do you think?

henriquesimoes (approved):

PR #88, opened 2024-11-29
base: remove repetition of environment variables.

ericonr:
Switch to using shell files defining the version for each module, and explicitly sourcing them in the relevant scripts.

This change makes adding new modules easier (2 fewer places to make changes), as well as simplifying adding any new variables for existing modules (e.g. checksums).


base/install_opcua.sh
```diff
@@ -3,6 +3,8 @@
 set -ex
 
 . /opt/epics/install-functions.sh
+. /opt/epics/epics_versions.sh
```
henriquesimoes, line 6: `$EPICS_BASE_PATH/epics_versions.sh`?

henriquesimoes (commented):
It indeed simplifies those definitions. Thanks!

I'm not very comfortable with the way we are copying those scripts to the image, though. How about placing them all in a known location beforehand (possibly outside `/opt`)?

base/install_opcua.sh
ericonr: We don't use `$EPICS_BASE_PATH` to source `install-functions.sh`, though. Should we do that? Otherwise it doesn't seem very consistent.

ericonr:
> I'm not very comfortable with the way we are copying those scripts to the image, though. How about placing them all in a known location beforehand

Is the issue copying them to the middle of the epics tree, or copying them for each install script?

Copying all of them at once loses the benefit of doing it for each stage, which is not invalidating the cache for packages which are built at the end.

henriquesimoes:
> Is the issue copying them to the middle of the epics tree

Yep. We are not very much consistent with the way run them. I'd rather have them in a known filesystem location and execute them with their absolute path than relying on the fact that they have been copied side-by-side somewhere in the epics tree.

Maybe placing them somewhere in the PATH, e.g. `/usr/local/bin`, is a sensible approach to avoid relying on yet another variable which defines where they are.

ericonr:
What about a single `/opt/epics-in-docker` (or similar) directory where everything goes? Otherwise we still have patches in the middle of the tree, and putting patches in `/usr/local/bin` feels wrong.

henriquesimoes:
Good point. I wasn't considering the patches. I'm okay with having them in `/opt/epics-in-docker`. ;)

ericonr:
> Good point. I wasn't considering the patches. I'm okay with having them in `/opt/epics-in-docker`. ;)

Don't forget to add it to the cleanup !

henriquesimoes (approved):
The tree in the resulting image is now clean. Thanks!

`$EPICS_RELEASE_FILE` could also be placed in `$EPICS_IN_DOCKER`, IMO. That's not a big deal, though.

PR #90, opened 2024-12-04
Add P4P (PVAGW)

ericonr:


base/modules_versions.sh
```diff
@@ -1,4 +1,5 @@
 PVXS_VERSION=1.3.1
+P4P_VERSION=69b1c282cece9f45ec13be23c6fc9a719e6503df # 4.2.0
```
ericonr, line 2: I will fix this in the next push: https://github.com/epics-base/p4p/issues/161

PR #93, opened 2024-12-11
ioc: allow pip to install packages system-wide.

henriquesimoes:
Since Debian 12, installing packages [outside a virtual environment is an error by default][1] as first specified by [PEP 668][2]. The rationale for forcing this is not applicable in our context, since the container image is dedicated to the IOC under construction, which renders it an isolated environment of its own.

Users should still provide RUNTIME_PIP_PACKAGES that are compatible with the system packages, not really breaking the system, similarly to what is done for shared libraries.

[1]: https://salsa.debian.org/python-team/packages/python-pip/-/commit/5aeba2464e36020a34e2bc074872771f5ef4973d
[2]: https://peps.python.org/pep-0668

Fixes: 39bb0d4e1d42 (base: update to Debian 12., 2024-10-31)


gustavosr8:
LGTM

gustavosr8 (approved):

ericonr (approved):

PR #95, opened 2024-12-13
LICENSE: add LICENSE file for BSD.

### LINKED ISSUE #94: Add LICENSE file to repository
Right now, this repository has no license defined, so by default [it is considered ARR](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository#choosing-the-right-license), which makes this project non-OSS. I assume this is not intended.

gustavosr8:


ericonr:
I don't know what it means to define build scripts as GPLv3.

I'd prefer using something closer to BSD, ISC or MIT, or even public domain.

gustavosr8:
ping @ericonr

LICENSE
```diff
@@ -0,0 +1,26 @@
+Copyright 2024 CNPEM
```
ericonr: 2023-2026 (assuming that fits the license format)

gustavosr8: Doesn't this means that the copyright is only valid till the end of 2026?
I've found other projects using this format, so assume it fits the format, but doesn't actually understand what does it means.

ericonr: It's usually about when changes happened. So you update the line when the year begins, and any new changes are under the same copyright.

ericonr (approved):

PR #97, opened 2024-12-17
ci: allow release branch to be protected.

ericonr:
The deploy key can be added to the bypass list for branch protection, allowing the release branch to be protected from pushes from regular users.

Based on:
https://github.com/orgs/community/discussions/25305#discussioncomment-10728028


henriquesimoes (approved):
I'm not sure how to properly review this. The `ssh-key` line seems fine. If the secret and rulesets have been configured, I think we are good to go.

PR #98, opened 2024-12-19
Add labels to the resulting image containing the build properties.

gustavosr8:


Dockerfile
```diff
@@ -1,9 +1,15 @@
 ARG DEBIAN_VERSION=12.8
+ARG EPICS_IN_DOCKER_VERSION=v0.12.0-dev
 
-FROM ghcr.io/cnpem/lnls-debian-epics-7:v0.12.0-dev AS build-image
+FROM ghcr.io/cnpem/lnls-debian-epics-7:${EPICS_IN_DOCKER_VERSION} AS build-image
```
ericonr, line 4: We might have to repeat these values in two places... The problem with making `EPICS_IN_DOCKER_VERSION` an argument is that it can be overridden by the `docker-compose.yml` definitions, which we don't want. We want the Dockerfile versions to be matched to the base image versions.

Dockerfile
```diff
@@ -1,9 +1,15 @@
...
 
 FROM debian:${DEBIAN_VERSION}-slim AS base
 
+ARG DEBIAN_VERSION
+LABEL org.opencontainers.image.debian_version=${DEBIAN_VERSION}
+ARG EPICS_IN_DOCKER_VERSION
+LABEL org.opencontainers.image.epics_in_docker_version=${EPICS_IN_DOCKER_VERSION}
```
ericonr, line 11: We should have our own namespace for labels. `br.lnls. ...` or `io.github.cnpem.epics-in-docker.version`

ericonr (changes requested):

Dockerfile
gustavosr8: I use as reference the base compose.
https://github.com/cnpem/epics-in-docker/blob/a9a761cd6a33da33c20cecb5fa5ab81d21f208fa/base/docker-compose.yml#L8-L10
Do you think we should also change there in order to have consistency on using our own namespace?

.github/workflows/base-image.yml
```diff
@@ -45,7 +45,7 @@ jobs:
             *.cache-to=type=gha,mode=max,scope=${{ matrix.image.name }}
           push: ${{ github.event_name == 'push' && vars.PUSH_TO_REGISTRY == 'true' && github.ref_type == 'tag' }}
       - name: Point IOC Dockerfile to image that was just built
-        run: sed -i "s/v.*-dev/$TAG/" Dockerfile
+        run: sed -i -E "s/(v[0-9]+\.[0-9]+\.[0-9]+-dev)/$TAG/" Dockerfile
```
henriquesimoes, line 48: Nit: we don't need the group matching here, that is, you can drop the parentheses around the regexp.

Dockerfile
```diff
@@ -80,6 +84,7 @@ FROM base AS dynamic-link
 
 COPY --from=dynamic-build /opt /opt
 
+LABEL org.opencontainers.image.build_target="dynamic-link"
```
henriquesimoes, line 87: Not a standard [Open Containers label](https://specs.opencontainers.org/image-spec/annotations/), right?

Dockerfile
```diff
@@ -94,6 +99,8 @@ RUN make distclean && make -j ${JOBS} && make $([ "$SKIP_TESTS" != 1 ] && echo r
 
 FROM base AS static-link
 
+LABEL org.opencontainers.image.link_type="static-link"
```
henriquesimoes, line 102: This is inconsistent with the other label.

henriquesimoes (changes requested):

Dockerfile
gustavosr8: Yep, missed this part when editing the commit, tnks!

Dockerfile
```diff
@@ -80,6 +84,7 @@ FROM base AS dynamic-link
 
 COPY --from=dynamic-build /opt /opt
 
+LABEL br.lnls.epics-in-docker.build-target="dynamic-link"
```
henriquesimoes, line 87: Maybe simply `target` is enough? There is no such thing as `runtime-target`, for instance.

henriquesimoes (changes requested):
Missing a label for the `no-build` target. BTW, is this really enough for identifying the resulting image? I think at least APP_DIRS will also be needed.

gustavosr8:
>  BTW, is this really enough for identifying the resulting image? I think at least APP_DIRS will also be needed.
Indeed. But first #59 needs to be merged to make sense. I will just convert this PR to a WIP and wait, for now.

.github/workflows/base-image.yml
```diff
@@ -45,7 +45,7 @@ jobs:
             *.cache-to=type=gha,mode=max,scope=${{ matrix.image.name }}
           push: ${{ github.event_name == 'push' && vars.PUSH_TO_REGISTRY == 'true' && github.ref_type == 'tag' }}
       - name: Point IOC Dockerfile to image that was just built
-        run: sed -i "s/v.*-dev/$TAG/" Dockerfile
+        run: sed -i -E "s/v[0-9]+\.[0-9]+\.[0-9]+-dev/$TAG/" Dockerfile
```
ericonr, line 48: The commit for this doesn't need to explain what the new sed expression does. It should instead explain why that was necessary.

Dockerfile
```diff
@@ -4,12 +4,17 @@ FROM ghcr.io/cnpem/lnls-debian-epics-7:v0.12.0-dev AS build-image
...
+ARG DEBIAN_VERSION
+
 ARG RUNDIR
 ARG ENTRYPOINT=/usr/local/bin/lnls-run
 ARG RUNTIME_PACKAGES
 ARG RUNTIME_TAR_PACKAGES
 ARG RUNTIME_PIP_PACKAGES
 
+LABEL br.lnls.epics-in-docker.version="v0.12.0-dev"
+LABEL br.lnls.debian.version=${DEBIAN_VERSION}
```
ericonr, line 16: this is missing an "epics-in-docker" level

ericonr (changes requested):

ericonr:
last commit message should be something like: "ioc: add labels with build properties to the IOC images."

henriquesimoes (changes requested):

ericonr:
I have reviewed, don't have any comments now, waiting on #19 being merged and this one rebased before actually approving.

Dockerfile
```diff
@@ -51,13 +56,17 @@ ARG APP_DIRS
 ARG RUNDIR
 ARG SKIP_PRUNE
 
+LABEL br.lnls.epics-in-docker.skip-prune=${SKIP_PRUNE}
+LABEL br.lnls.epics-in-docker.app-dirs=${APP_DIRS}
```
henriquesimoes, line 60: These should be populated when using the `dynamic-link` target as well.

henriquesimoes (changes requested):

Dockerfile
```diff
@@ -51,13 +56,17 @@ ARG APP_DIRS
 ARG RUNDIR
```
henriquesimoes, line 56: Considering `RUNDIR` is also an argument to `lnls-prune-artifacts`, it makes sense to call it a build properly as well. What do you think?

gustavosr8: yeap, seems relevant to me!

henriquesimoes (approved):
I missed a simple note in the commit message about why having labels is interesting after all, and the rationale for including some variables and not the others.

We all know because we have discussed this here, but future readers deserve to know as well.

Dockerfile
```diff
@@ -87,15 +97,21 @@ ARG RUNDIR
 ARG SKIP_TESTS
 ARG SKIP_PRUNE
 
+
+LABEL br.lnls.epics-in-docker.skip-prune=${SKIP_PRUNE}
+LABEL br.lnls.epics-in-docker.app-dirs=${APP_DIRS}
+LABEL br.lnls.epics-in-docker.rundir=${RUNDIR}
```
henriquesimoes, line 103: I've tested with ad-sim-epics-ioc, and this approach actually doesn't work. :/

When we COPY part of the filesystem, these labels won't be inherited. We must define them in the `*-link` stages.

henriquesimoes (changes requested):
I should've tested before submitting my review ;)

Better later than never.

Dockerfile
```diff
@@ -13,6 +15,9 @@ ARG RUNTIME_PIP_PACKAGES
 ARG PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 ENV PATH=$PATH
 
+LABEL br.lnls.epics-in-docker.version="v0.12.0-dev"
```
ericonr, line 18: This duplication makes me a bit sad. What could we do to avoid it?

Should we have a release script to automate changing these values? Does `ENV` work for image names (I don't think `ENV` can be overridden during builds)?

henriquesimoes: > Does ENV work for image names (I don't think ENV can be overridden during builds)?

It won't work for image name and tag. It must be an ARG for that: ["ARG is the only instruction that may precede FROM in the Dockerfile."](https://docs.docker.com/reference/dockerfile/#from)

Unfortunately, this would reopen the door we closed back in v0.3.0 (d240ed0).

> Should we have a release script to automate changing these values?

Maybe that's the only option there is left. I'll launch a background thread in my mind, and see if any idea comes up.

henriquesimoes: @gustavosr8, you must updated it to the latest development version (v0.13.0-dev) manually if we cannot make this happen automatically.

gustavosr8: > Should we have a release script to automate changing these values?

The idea is to make a text-replacement script to change this value in Dockerfile? If we go on this direction, we will have a script committing in the repo, an it doesn't sounds very pleasant to me

ericonr: > The idea is to make a text-replacement script to change this value in Dockerfile? If we go on this direction, we will have a script committing in the repo, an it doesn't sounds very pleasant to me

Essentially that. Most of the procedure in `RELEASE.md` could be automated, and it would only commit, there would still be the need for a PR to review the resulting commits. It could even open `CHANGES.md` in an editor for you in order to compose the release notes...

gustavosr8: Hmm I see...

> Most of the procedure in `RELEASE.md` could be automated

What about make it a new project/issue then? I think that include this here will be out of the scope of this PR

ericonr: > I think that include this here will be out of the scope of this PR

Out of scope, yes, but I don't think we should merge this until we have the script.

gustavosr8:
@ericonr @henriquesimoes I think we can return to this now :sweat_smile:

Dockerfile
```diff
@@ -87,15 +100,24 @@ ARG RUNDIR
 ARG SKIP_TESTS
 ARG SKIP_PRUNE
 
+
```
ericonr, line 103: accidental newline

Dockerfile
```diff
@@ -87,15 +100,24 @@ ARG RUNDIR
...
 FROM base AS dynamic-link
 
-COPY --from=dynamic-build /opt /opt
+ARG RUNDIR
+ARG SKIP_TESTS
+ARG SKIP_PRUNE
+
+LABEL br.lnls.epics-in-docker.skip-prune=${SKIP_PRUNE}
+LABEL br.lnls.epics-in-docker.app-dirs=${APP_DIRS}
+LABEL br.lnls.epics-in-docker.rundir=${RUNDIR}
```
ericonr, line 117: `SKIP_TESTS` wasn't used, and `ADD_DIRS` isn't an `ARG`.

I think the order of args and labels should match, too

ericonr (changes requested):

Dockerfile
```diff
@@ -56,8 +61,16 @@ RUN if [ "$SKIP_PRUNE" != 1 ]; then lnls-prune-artifacts ${APP_DIRS} ${RUNDIR};
...
+ARG SKIP_PRUNE
+ARG APP_DIRS
+ARG RUNDIR
 
+LABEL br.lnls.epics-in-docker.skip-prune=${SKIP_PRUNE}
+LABEL br.lnls.epics-in-docker.app-dirs=${APP_DIRS}
+LABEL br.lnls.epics-in-docker.rundir=${RUNDIR}
+LABEL br.lnls.epics-in-docker.target="no-build"
+
+COPY --from=pruned-build /opt /opt
```
ericonr, line 74: Deleted newline separating stages

Dockerfile
```diff
@@ -94,8 +107,16 @@ RUN if [ "$SKIP_PRUNE" != 1 ]; then lnls-prune-artifacts ${APP_DIRS} ${PWD} ${RU
...
+ARG SKIP_PRUNE
+ARG APP_DIRS
+ARG RUNDIR
 
+LABEL br.lnls.epics-in-docker.skip-prune=${SKIP_PRUNE}
+LABEL br.lnls.epics-in-docker.app-dirs=${APP_DIRS}
+LABEL br.lnls.epics-in-docker.rundir=${RUNDIR}
+LABEL br.lnls.epics-in-docker.target="dynamic-link"
+
+COPY --from=dynamic-build /opt /opt
```
ericonr, line 120: Deleted newline separating stages

ericonr (changes requested):

ericonr (changes requested):
The newlines are still missing

we should always have two empty lines before a `FROM` statement defining a new build stage.

gustavosr8:
> The newlines are still missing
 Oh sorry, I've read "Delete" instead of "Deleted". My bad :sweat_smile:

ericonr (approved):

PR #101, opened 2025-02-17
Add CA Gateway image.

ericonr:
I used the release versions for PCAS and the gateway itself. The only fixes we might be missing from the master branch would be:

https://github.com/epics-modules/pcas/commit/a8ba939c3313cb634cdb9ab751fca53d5080e00e <-- if using addr list, use it for beacons as well (I'm not sure it's relevant to us)

https://github.com/epics-extensions/ca-gateway/commit/ec541d17c50327989adc29ecb5a7a99a7b990106 <-- performance improvement for arrays

I'm pending towards leaving as is, but I'm open to suggestions.


base/install_cagateway.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -ex
+
+. $EPICS_IN_DOCKER/install-functions.sh
+. $EPICS_IN_DOCKER/modules_versions.sh
```
henriquesimoes, line 6: `s/modules_versions.sh/cagateway_versions.sh/`.

That should fix CI, I hope. Maybe we should (re?)consider using `set -u`. It would have made this mistake easier to catch.

ericonr: Eh, I hadn't even built locally, so I don't think that was the only thing missing.

images/docker-compose-ca-gateway.yml
```diff
@@ -0,0 +1,13 @@
+services:
+  pvagw:
```
henriquesimoes, line 2: Wrong service name.

ericonr:
The image is now built in CI, and I can launch the gateway inside it.

henriquesimoes (approved):

PR #103, opened 2025-02-25
base: speed up local rebuilds.

### LINKED ISSUE #60: Evaluate ccache and a bind volume to speed up offline base builds

ericonr:
Using ccache can speed up local rebuilds even when the container layer cache is invalidated, which is easy to happen. It also requires a less complicated and less fragile setup than distcc (though distcc will always make for faster first builds).

install_epics.sh still supports ccache-less setups, should that be required.

On my machine, building an image with ccache for the first time took ~15m. Rebuilding it after invalidating the layer cache took only 3m36s. The output of 'ccache -s' is reproduced below:

    Cacheable calls:   3858 / 8743 (44.13%)
      Hits:            1923 / 3858 (49.84%)
        Direct:        1822 / 1923 (94.75%)
        Preprocessed:   101 / 1923 ( 5.25%)
      Misses:          1935 / 3858 (50.16%)
    Uncacheable calls: 4885 / 8743 (55.87%)
    Local storage:
      Cache size (GB): 0.18 / 5.00 ( 3.70%)

It is not clear why the cache hits were below 50%, but it was still enough to make a big difference in build times.

This was based on StackOverflow [1].

[1] https://stackoverflow.com/a/56833198

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -5,6 +5,8 @@ FROM debian:${DEBIAN_VERSION}
 ARG JOBS
 
 ENV DEBIAN_FRONTEND=noninteractive
+ARG CCACHE_DIR=/ccache
+ARG CCACHE_STATSLOG=/tmp/ccache.statslog
 
 # Disable APT sandbox so that single UID restriction is satisfied in systems
 # without subuid and subgid configured.
@@ -14,6 +16,7 @@ RUN apt update -y && \
     apt install -y --no-install-recommends \
         build-essential \
         cmake \
+        ccache \
         git \
         libaravis-dev \
         libboost-test-dev \
@@ -37,7 +40,8 @@ RUN apt update -y && \
         cython3 \
         python3-numpy \
         python3-setuptools \
-        ca-certificates
+        ca-certificates && \
+    for p in gcc g++; do ln -vs /usr/bin/ccache /usr/local/bin/$p; done
 
 COPY lnls-get-n-unpack.sh /usr/local/bin/lnls-get-n-unpack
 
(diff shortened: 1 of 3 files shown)


ericonr:
I'm using `ARG` for `CCACHE_DIR`, but maybe I should use `ENV`, so building IOC images still uses a well defined path instead of the `$HOME/.config/ccache` default (based on <https://ccache.dev/manual/4.10.2.html#_configuration_options>).

Having those files in the build image shouldn't be an issue anyway, since all our images use multi-stage builds with copies into a clean image.

henriquesimoes:
> I'm using ARG for CCACHE_DIR, but maybe I should use ENV

In fact, it doesn't make much sense to have it as an `ARG` if `--mount` won't have any different path than the one predefined by us.

> Having those files in the build image shouldn't be an issue anyway, since all our images use multi-stage builds with copies into a clean image.

I'm not sure I grasped which files you're referring to. Are you talking about not defining `ENV` at all and letting it write to `$HOME/.config/ccache`?

ericonr:
Yay, it built. Will test the gateway here to see if everything's fine.

base/musl/Dockerfile
```diff
@@ -4,8 +4,11 @@ FROM alpine:$ALPINE_VERSION
 
 ARG JOBS
 
-RUN apk add --no-cache \
+ARG CCACHE_DIR=/ccache
```
henriquesimoes, line 7: It works just fine to have this as an `ENV`. I'd rather have it this way than an `ARG`.

Alternatively, we can go with the default path ccache uses. It seems shorter and easier to simply define `CCACHE_DIR`, though.

henriquesimoes (changes requested):
@gustavosr8 is willing to merge this sooner, as our build time is increasing significantly over time.

base/musl/Dockerfile
```diff
@@ -55,11 +60,14 @@ WORKDIR /opt/epics
...
+RUN --mount=type=cache,target=/ccache/ USE_CCACHE=1 COMMANDLINE_LIBRARY=READLINE_NCURSES $EPICS_IN_DOCKER/install_epics.sh
 
 WORKDIR ${EPICS_MODULES_PATH}
 
 COPY modules_versions.sh install_modules.sh $EPICS_IN_DOCKER
-RUN NEEDS_TIRPC=YES $EPICS_IN_DOCKER/install_modules.sh
+RUN --mount=type=cache,target=/ccache/ NEEDS_TIRPC=YES $EPICS_IN_DOCKER/install_modules.sh
+
+RUN ccache --show-log-stats && \
+    for p in gcc g++; do ln -vsf /usr/bin/$p /usr/local/bin/$p; done
```
henriquesimoes, line 71: Can you highlight in the commit message we can't simply `unlink` them because we won't revert the `CONFIG.gnuCommon` from epics-base?

base/Dockerfile
```diff
@@ -37,7 +40,9 @@ RUN apt update -y && \
         cython3 \
         python3-numpy \
         python3-setuptools \
-        ca-certificates
+        ca-certificates && \
+    ccache -z && \
```
henriquesimoes: Leftover `ccache -z`

henriquesimoes (approved):
LGTM. Thanks!

PR #105, opened 2025-03-07
Distro updates: Debian 12.9 and Alpine 3.20.6

ericonr:


ericonr:
With Alpine 3.21.3, `py3-numpy-dev` no longer automatically installs `py3-numpy`... It seems to work on `edge` though, so was a temporary bug. Will see if I open an issue, or we simply skip 3.21...

For now, stick to updating to latest of the 3.20 series.

PR #106, opened 2025-03-07
Release 0.13.0

ericonr:


CHANGES.md
```diff
@@ -2,6 +2,17 @@
...
+## v0.13.0
+
+Users of the `dynamic-link` target for IOC images should update in order to
+take advantage of the new image pruning steps, which should greatly lower the
+size of those IOC images.
+
+Users interested in improvements added in EPICS 7.0.8 should also update.
+
+This release includes 3 new images: one with epics-base and PVXS tools, and one
+for each EPICS gateway.
```
henriquesimoes, line 14: In previous releases, we also highlighted the new modules users might be interested in. In this case, it would include p4p (and pcas if we were to recommend it to anyone, which I'm afraid we're not).

gustavosr8 (approved):

CHANGES.md
ericonr: p4p is not a module to be used by IOCs, it's there mainly because it's where the PVAGW implementation lives.

I don't think we want to recommend epics-in-docker as a base for projects which use p4p. What do you think?

henriquesimoes: I raised this question exactly to get to know if it we were forgetting to mention it, or if it doesn't make sense at all. :)

I don't know the p4p project myself to properly judge this. If it doesn't provide any interface to be used by a softIOC in Python, I'm okay to consider it a module not supported by epics-in-docker (although we do ship it in the end).

ericonr: As discussed offline, we don't intend to support using p4p in IOCs with something like pydevsup. Furthermore, the decision to not list p4p in the changelog was made when adding the PR to the changelog, so it will remain as such.

PR #112, opened 2025-03-11
RELEASE: automate the release process.

gustavosr8:
This script automates the process described in RELEASE.md, pushing all the modifications to a new branch called release-$RELEASE_VERSION. One must open a PR to main after completing all the steps.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -9,46 +9,30 @@ necessary.
 
 Make sure there aren't any pending issues before tagging a new release.
 
-## Release commit
+## Run release script
 
-### Change version in `Dockerfile`
+The release script is used to make the necessary changes to enable tagging
+version `vX.Y.Z`:
 
-The image version specified in [`Dockerfile`](./Dockerfile) should be `vX.Y.Z`.
+```
+$ ./make-release.sh vX.Y.Z
+```
 
-### Change version in CHANGES
+The script will create a separate branch, where it will make the necessary
+changes and prompt you for any further required information.
 
-The "Unreleased" version in [CHANGES](./CHANGES.md) should be changed to
-`vX.Y.Z`, and a new "Unreleased" header should be added.
-
-### Add release description to CHANGES
-
-This description should answer the "who needs this update?" question.
-
-### Create release commit
-
-This commit should contain only the above change, and be titled `Release
-version vX.Y.Z.`.
-
-### (Optional) Create Pull Request for this commit
+### (Optional) Create Pull Request for these commits
 
 Make sure the description covers all relevant points.
 
 ### Create tag
 
-The tag should be signed and created locally:
+The tag should be signed and created locally, and point to the release commit
+which was merged to the main branch:
 
 ```
-$ git tag -a -s vX.Y.Z
+$ git tag -a -s -m vX.Y.Z vX.Y.Z <release-commit-ref>
 ```
-
-And its contents should be only the title: `vX.Y.Z`.
 
 For this step, it is necessary to [configure Git to sign tags with SSH
 keys](https://docs.gitlab.com/ee/user/project/repository/signed_commits/ssh.html).
-
-## `main` branch
-
-A new commit should be created in the `main` branch, updating the version used
-in [Dockerfile](./Dockerfile) to `vX.Y.Z-dev` to avoid usage of the `main`
-branch in actual deployments. It should be titled `Post-release development
-version bump.`.
--- a/make-release.sh
+++ b/make-release.sh
@@ -1,0 +1,32 @@
+#!/bin/sh
+
+set -e
+
+if [ -z "$1" ]; then
+    echo "Usage: $0 <release_version>"
+    exit 1
+fi
+
+RELEASE_VERSION=$1
+
+# Update CHANGES.md
+sed -i "s/## Unreleased/&\n\n## $RELEASE_VERSION\n\n\
+% RELEASE NOTES TEMPLATE - Lines starting with '%' will be ignored.\n\
+% This description should answer \"who needs this update?\"\n\
+% Please start your description in the empty line below, where your cursor is, without adding new empty lines.\n/" CHANGES.md
+${EDITOR:-vim} +10 CHANGES.md
+sed -i '/^%/d' CHANGES.md
+
+# Update version in Dockerfile
+sed -i -E "s/v[0-9]+\.[0-9]+\.[0-9]+-dev/$RELEASE_VERSION/" Dockerfile
+
+git switch -c release-$RELEASE_VERSION
+git commit -m "Release version ${RELEASE_VERSION}." CHANGES.md Dockerfile
+
+# Post-release version bump
+sed -i -E "s/$RELEASE_VERSION/&-dev/" Dockerfile
+
+git commit -m "Post-release development version bump." Dockerfile
+
+echo "Check your local modifications before submitting with"
+echo "git push origin release-$RELEASE_VERSION"


make-release.sh
```diff
@@ -0,0 +1,35 @@
...
+
+if [ -z "$1" ]; then
+    echo "Usage: $0 <release_version>"
+    exit 1
+fi
+
+RELEASE_VERSION=$1
+
+# Update CHANGES.md
+sed -i "s/## Unreleased/## Unreleased\n\n## $RELEASE_VERSION\n/" CHANGES.md
```
henriquesimoes, line 11: You can use `&` to replicate the match in the output:

```bash
s/## Unreleased\n/&\n## $RELEASE_VERSION\n/
```

make-release.sh
```diff
@@ -0,0 +1,35 @@
...
+echo "This description should answer the \"who needs this update?\" question."
+sleep 10
+${EDITOR:-vim} +7 CHANGES.md
```
henriquesimoes, line 14: How about writing guide line in the CHANGES.md file itself? No need to sleep this way.

You can make something to make the line stand out, to avoid being forgotten.

henriquesimoes (commented):
Partial review. Still need to walk to the last of the scripts after lunch. ;)

make-release.sh
gustavosr8: `sed` needs any extra flag in order to this work?

make-release.sh
```diff
@@ -0,0 +1,35 @@
...
+# Update version in Dockerfile
+sed -i "s/FROM ghcr\.io\/cnpem\/lnls-debian-epics-7:.*/FROM ghcr.io\/cnpem\/lnls-debian-epics-7:$RELEASE_VERSION AS build-image/" Dockerfile
+
+git checkout -b release-$RELEASE_VERSION
+git add CHANGES.md Dockerfile
+git commit -m "Release version ${RELEASE_VERSION}."
+
+# Create signed tag
+git tag -a -s $RELEASE_VERSION -m "$RELEASE_VERSION"
+git push origin release-$RELEASE_VERSION
```
henriquesimoes, line 25: I'd rather have this done manually.

Even if we do want to push automatically, this shouldn't need to be done twice (once here and again after the post-release commit).

make-release.sh
```diff
@@ -0,0 +1,35 @@
...
+git push origin $RELEASE_VERSION
```
henriquesimoes, line 26: We cannot do this until the PR is approved. Otherwise, the tag will trigger a CI that will build and push the images.

make-release.sh
```diff
@@ -0,0 +1,35 @@
...
+RELEASE_VERSION=$1
+
+# Update CHANGES.md
+sed -i "s/## Unreleased/## Unreleased\n\n## $RELEASE_VERSION\n/" CHANGES.md
+echo "This description should answer the \"who needs this update?\" question."
+sleep 10
+${EDITOR:-vim} +7 CHANGES.md
+
+# Update version in Dockerfile
+sed -i "s/FROM ghcr\.io\/cnpem\/lnls-debian-epics-7:.*/FROM ghcr.io\/cnpem\/lnls-debian-epics-7:$RELEASE_VERSION AS build-image/" Dockerfile
```
henriquesimoes, line 17: This looks like we will still need to keep in sync where version appears.

Can't we do something [similar to the CI job](https://github.com/cnpem/epics-in-docker/pull/98/commits/24101489941fb973b78fb6c51d1775a4caf57181)?

henriquesimoes (changes requested):
IMHO, everything should be done locally only. Pushing to the remote repository should be done after the author itself checking if everything went as expected.

make-release.sh
henriquesimoes: I broke my example by moving the `\n` to the match after testing it. I forgot `sed` tokenizes based on lines. Sorry. Try this instead:

```
sed "s/## Unreleased/&\n\n## $RELEASE_VERSION\n/"
```

gustavosr8: Now it work, thanks :)

gustavosr8:
> IMHO, everything should be done locally only. Pushing to the remote repository should be done after the author itself checked if everything went as expected.

Ok, so no automatic push, but what about the other git stuff (checkout to a new branch, stage files, commit and tag)?

henriquesimoes:
> what about the other git stuff (checkout to a new branch, stage files, commit and tag)?

I have no problem with those. They are easier to clean up with anything goes wrong. Writing to the remote server seems out of the script scope. It would fail if we don't Internet access, for instance, even though it accomplished the most painful task which it aims at solving.

gustavosr8:
@henriquesimoes seems fair to me. Will just add a reminder to push the local modifications at the end.

make-release.sh
```diff
@@ -0,0 +1,35 @@
...
+sed -i "s/## Unreleased/&\n\n## $RELEASE_VERSION\n\n%RELEASE NOTES TEMPLATE - DON'T EDIT THE LINES STARTING WITH % \n%This description should answer the \"who needs this update?\" question.\n/" CHANGES.md
+${EDITOR:-vim} +9 CHANGES.md
+sed -i '/^%/d' CHANGES.md
+
+# Update version in Dockerfile
+sed -i -E "s/v[0-9]+\.[0-9]+\.[0-9]+-dev/$RELEASE_VERSION/" Dockerfile
+
+git checkout -b release-$RELEASE_VERSION
+git add CHANGES.md Dockerfile
+git commit -m "Release version ${RELEASE_VERSION}."
```
ericonr, line 20: You can use `git commit -m  "title" <files>`, so the commit only contains those files, instead of whatever else could be in index.

make-release.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# Create signed tag
+git tag -a -s $RELEASE_VERSION -m "$RELEASE_VERSION"
```
ericonr, line 23: Have the switch argument `-m "$RELEASE_VERSION"` before the positional argument

make-release.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# Post-release version bump
+DEV_VERSION="${RELEASE_VERSION}-dev"
+
+sed -i "s/FROM ghcr\.io\/cnpem\/lnls-debian-epics-7:.*/FROM ghcr.io\/cnpem\/lnls-debian-epics-7:$DEV_VERSION AS build-image/" Dockerfile
+sed -i -E "s/v[0-9]+\.[0-9]+\.[0-9]+/$DEV_VERSION/" Dockerfile
```
ericonr, line 29: Why isn't the second sed enough?

make-release.sh
```diff
@@ -0,0 +1,35 @@
...
+    echo "Usage: $0 <release_version>"
+    exit 1
+fi
+
+RELEASE_VERSION=$1
+
+# Update CHANGES.md
+sed -i "s/## Unreleased/&\n\n## $RELEASE_VERSION\n\n%RELEASE NOTES TEMPLATE - DON'T EDIT THE LINES STARTING WITH % \n%This description should answer the \"who needs this update?\" question.\n/" CHANGES.md
+${EDITOR:-vim} +9 CHANGES.md
+sed -i '/^%/d' CHANGES.md
```
ericonr, line 13: Why can't they edit lines with `%`? Wouldn't removing them simply make the later sed a no-op?

ericonr (commented):
I just realized that maybe tagging shouldn't be part of the script?

If we are going to open a PR, approve it, and merge it via the web UI, the resulting commits in `main` won't be the same hash. The tag has to be created after merging the PR and pulling the latest version of main :thinking: 

Maybe the script shouldn't do any tagging?

make-release.sh
gustavosr8: Oh, didn't know that you can do it. Thanks!

make-release.sh
gustavosr8: It is. Forgot this line in the previous modifications 😅

gustavosr8:
> The tag has to be created after merging the PR and pulling the latest version of main 🤔
Hmm, hadn't thought about it... but makes sense to me!


Is there any way we can reinforce the need for it, assuming one will not read RELEASE.md since the script already follows the steps for them?

henriquesimoes:
> If we are going to open a PR, approve it, and merge it via the web UI, the resulting commits in main won't be the same hash.

How about making this an exceptional case, and manually fast-forward merging? Is that something too off to make it part of the procedure?

> one will not read RELEASE.md since the script already follows the steps for them?

Talking about that, shouldn't the script itself be the source of truth from now on? That is, shouldn't we document in RELEASE.md how to use the script instead of documenting the steps the script does?

gustavosr8:
> How about making this an exceptional case, and manually fast-forward merging? Is that something too off to make it part of the procedure?

Since we already use the rebase and merge method instead of 3 way, I don't see any downside on this!

> shouldn't the script itself be the source of truth from now on? 

Well, I think it's better to have the opinion of the document's author here (@ericonr 👀)

gustavosr8:
So, as we discussed in person, the scope of this project is to make the text modifications, write the commits as defined in RELEASE.md document and create the release branch with this modifications. Now, the document should describe the script usage and further steps for complete the release process.

RELEASE.md
```diff
@@ -9,33 +9,22 @@ necessary.
 
 Make sure there aren't any pending issues before tagging a new release.
 
-## Release commit
+## Release script
```
ericonr, line 13: Run release script

RELEASE.md
```diff
@@ -9,33 +9,22 @@ necessary.
...
 
-### Change version in `Dockerfile`
+Run `make-release.sh vX.Y.Z`. The script will change the image version
+specified in [`Dockerfile`](./Dockerfile), change the version in
+[CHANGES](./CHANGES.md) and add a release description, create a release commit
+in a new release branch and update the version used in
+[Dockerfile](./Dockerfile) to `vX.Y.Z-dev` to avoid usage of the `main` branch
+in actual deployments.
```
ericonr, line 19: The release script is used to make the necessary changes to enable tagging version `vX.Y.Z`:

```
$ ./make-release.sh vX.Y.Z
```

The script will create a separate branch, where it will make the necessary changes and prompt you for any further required information.

RELEASE.md
```diff
@@ -9,33 +9,22 @@ necessary.
...
-
-This description should answer the "who needs this update?" question.
-
-### Create release commit
-
-This commit should contain only the above change, and be titled `Release
-version vX.Y.Z.`.
-
-### (Optional) Create Pull Request for this commit
+### (Optional) Create Pull Request for this commits
```
ericonr, line 21: these

RELEASE.md
```diff
@@ -9,33 +9,22 @@ necessary.
...
 
 Make sure the description covers all relevant points.
 
 ### Create tag
 
-The tag should be signed and created locally:
+The tag should be signed and created locally, and point to the release commit:
```
ericonr, line 27: to the release commit which was merged to the main branch

RELEASE.md
```diff
@@ -9,33 +9,22 @@ necessary.
...
 
 ```
 $ git tag -a -s vX.Y.Z
```
ericonr, line 30: `git tag -a -s -m vX.Y.Z vX.Y.Z <release-commit-ref>`

make-release.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
ericonr, line 1: `/usr/bin/env bash`

make-release.sh
```diff
@@ -0,0 +1,31 @@
...
+sed -i -E "s/v[0-9]+\.[0-9]+\.[0-9]+-dev/$RELEASE_VERSION/" Dockerfile
+
+git switch -c release-$RELEASE_VERSION
+git add CHANGES.md Dockerfile
+git commit -m "Release version ${RELEASE_VERSION}." CHANGES.md Dockerfile
+
+# Post-release version bump
+DEV_VERSION="${RELEASE_VERSION}-dev"
+
+sed -i -E "s/v[0-9]+\.[0-9]+\.[0-9]+/$DEV_VERSION/" Dockerfile
```
ericonr, line 25: We know it's `RELEASE_VERSION`, so use that

make-release.sh
```diff
@@ -0,0 +1,31 @@
...
+
+git add Dockerfile
+git commit -m "Post-release development version bump."
```
ericonr, line 28: Same `git commit` thing here, don't use `add`

ericonr (changes requested):
Commit message nit:

> This script automates the process described in RELEASE.md, pushing all
the modifications to a new branch called release-$RELEASE_VERSION. One
must open a PR to main after completing all the steps.

This script automates the process described in RELEASE.md, making all
the modifications to a new branch called release-$RELEASE_VERSION. One
must open a PR to main after completing all the steps.

make-release.sh
```diff
@@ -0,0 +1,31 @@
...
+git switch -c release-$RELEASE_VERSION
+git add CHANGES.md Dockerfile
+git commit -m "Release version ${RELEASE_VERSION}." CHANGES.md Dockerfile
+
+# Post-release version bump
+DEV_VERSION="${RELEASE_VERSION}-dev"
+
+sed -i -E "s/$RELEASE_VERSION/$DEV_VERSION/" Dockerfile
+
+git add Dockerfile
```
ericonr, line 27: The `git add` commands aren't necessary anymore..

make-release.sh
```diff
@@ -0,0 +1,31 @@
...
+
+if [ -z "$1" ]; then
+    echo "Usage: $0 <release_version>"
+    exit 1
+fi
+
+RELEASE_VERSION=$1
+
+# Update CHANGES.md
+sed -i "s/## Unreleased/&\n\n## $RELEASE_VERSION\n\n%RELEASE NOTES TEMPLATE - Lines starting with '%' will be ignored.\n%This description should answer the \"who needs this update?\" question.\n/" CHANGES.md
```
ericonr, line 11: This description should answer "who needs this update?"

Can you try splitting the sed argument into multiple lines?

ericonr (changes requested):

make-release.sh
```diff
@@ -0,0 +1,31 @@
...
+sed -i -E "s/v[0-9]+\.[0-9]+\.[0-9]+-dev/$RELEASE_VERSION/" Dockerfile
+
+git switch -c release-$RELEASE_VERSION
+git add CHANGES.md Dockerfile
+git commit -m "Release version ${RELEASE_VERSION}." CHANGES.md Dockerfile
+
+# Post-release version bump
+DEV_VERSION="${RELEASE_VERSION}-dev"
+
+sed -i -E "s/$RELEASE_VERSION/$DEV_VERSION/" Dockerfile
```
henriquesimoes, line 25: We can make this simpler: `s/$RELEASE_VERSION/&-dev/`, and get rid of `DEV_VERSION`.

make-release.sh
henriquesimoes, line 11: An empty space after `%` would be nice to improve readability.

RELEASE.md
```diff
@@ -9,46 +9,31 @@ necessary.
...
+```
+$ ./make-release.sh vX.Y.Z
+```
 
-### Create release commit
+The script will create a separate branch, where it will make the necessary
+changes and prompt you for any further required information.
 
-This commit should contain only the above change, and be titled `Release
-version vX.Y.Z.`.
```
henriquesimoes, line 24: Nit: we should remove one of the newlines surrounding this paragraph. Otherwise, it ends up two empty lines in a sequel.

henriquesimoes (approved):
Just a small formatting leftover bit now. Apart from that, it looks nice to me. Thanks!

make-release.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
gustavosr8: Seems fair to me!

make-release.sh
```diff
@@ -0,0 +1,32 @@
...
+    exit 1
+fi
+
+RELEASE_VERSION=$1
+
+# Update CHANGES.md
+sed -i "s/## Unreleased/&\n\n## $RELEASE_VERSION\n\n\
+% RELEASE NOTES TEMPLATE - Lines starting with '%' will be ignored.\n\
+% This description should answer \"who needs this update?\"\n\
+% Please enter your description right below, with no empty lines.\n/" CHANGES.md
```
ericonr, line 16: Please start your description in the empty line below, without adding new empty lines

ericonr: It was also necessary to update the offset passed to the editor

ericonr: I did the two fixes myself.

The resulting line was a bit longer than I'd like it; if you are comfortable with the size, feel free to merge this :)

ericonr (approved):

PR #113, opened 2025-03-20
ioc: defer git directory removal until after build.

henriquesimoes:
The .git directory is removed mainly for reducing the size of the resulting IOC image. However, doing this before compiling the IOC rules out the possibility of automatically detecting the version during the build (e.g. by calling `git describe` in the Makefile). Defer the removal to right after the IOC has been built, but still before it is COPY'ed to the final image, satisfying both goals.

The downside of this approach is the duplication in both dynamic and static build stages, which need to be kept in sync. Given that a similar issue already occurs for the build pipeline itself and that the benefits outweigh the drawbacks, it seems fine by now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -90,7 +90,6 @@ WORKDIR /opt/${REPONAME}
 COPY . .
 
 RUN cp $EPICS_RELEASE_FILE configure/RELEASE
-RUN rm -rf .git/
 
 
 FROM build-stage AS dynamic-build
--- a/Dockerfile
+++ b/Dockerfile
@@ -101,7 +101,7 @@ ARG RUNDIR
 ARG SKIP_TESTS
 ARG SKIP_PRUNE
 
-RUN make distclean && make -j ${JOBS} && make $([ "$SKIP_TESTS" != 1 ] && echo runtests) && make clean && make -C ${RUNDIR}
+RUN lnls-build-ioc
 
 RUN if [ "$SKIP_PRUNE" != 1 ]; then lnls-prune-artifacts ${APP_DIRS} ${PWD} ${RUNDIR}; fi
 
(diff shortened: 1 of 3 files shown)


gustavosr8:
nit: what about rewrite the commit as "ioc: postpone removal of git directory until after build". Sounds better to me!

ericonr:
> Given that a similar issue already occurs for the build pipeline itself 

If there's already one line that's duplicated, what about including this command in that line (so `&& rm -rf .git` instead of another `RUN` step)? I think that would make it easier to keep in sync, but it does make for quite a long line :)

For either one of the solutions we choose, I think we should also leave a comment on top of both lines specifying that it should be kept the same as the other target's line.

henriquesimoes:
> If there's already one line that's duplicated, what about including this command in that line (so `&& rm -rf .git` instead of another RUN step)?

I considered that for an instant as well. But since it doesn't change the layers storage AFAIU, I let alone for simplicity. We can follow that approach to keep them even closer together.

Another thing I did consider for an instant was to extract all those steps into an script and run it in both steps. It would be necessarily in sync. However, it does introduces a redirection for a rather simple set of operations. Is that worth the cost? If so, I'm happy to introduce a cleanup patch for that. If not, I'll apply your both suggestions ;)

ericonr:
> I think this answer could be thought of as: are we going to make other common changes to both dynamic and static build stages?

I think so!

There was a plan at one point of forcing all IOCs built with epics-in-docker to include reccaster and PVAccess (though not clear if using PVXS or implementation in epics-base).

To move in that direction, we'd need a standard way of building IOCs, probably involving a script to be called to set up the build with the correct options. And that's yet another step to add to both targets.

So maybe it makes sense to create a script?

PR #117, opened 2025-06-25
base: add lakeshore module

guirodrigueslima:
The lakeshore336 directory can be build as a standard EPICS support module.


gustavosr8 (approved):

ericonr:
Sorry for the delay, I was going to leave a comment now. @guirodrigueslima did you manage to test an IOC image that pulls the protocol from this module? You'll probably have to use a full path for it.

https://github.com/ISISComputingGroup/EPICS-lakeshore/blob/master/lakeshore336/protocol/Makefile uses a custom `DATA +=` directive in the Makefile, which doesn't exist for us, meaning the file isn't installed anywhere.

We could try to add support to it to our epics-base version, or do what epics-containers does:

> Our container build discovers them and puts them in a fixed location and adds a single fixed path to STREAM_PROTOCOL_PATH.

But the best path forward (from previous discussions) seems to be adding `DB += .proto` for protocol files; that way they are properly installed into `module/db`. This was a patch I would have liked to include into this module. For now, I think we can live with this version, but I will make a PR to the repository to install them using `DB`, and then later we can update the resulting IOC to use a better path.

PR #120, opened 2025-07-16
base: update motorParker version and fix bug in lakeshore module

guirodrigueslima:
The motorParker module version was updated to R1-1-1-oem-3. In the lakeshore336 module, it was observed that the folder was not removed during the image build process. The solution was to move the subfolder from EPICS-lakeshore/lakeshore336 to lakeshore336.


gustavosr8:
I would also write in the commit message what features of the new motorParker version are we interested on to justify the update.

ericonr:
> I would also write in the commit message what features of the new motorParker version are we interested on to justify the update.

Indeed!

We already know

> The motorParker module version was updated to R1-1-1-oem-3.

from the diff, the main information is what's changed. From what I could see, you made some changes to the base commit adding support for the OEM controller. Is this module in active development/testing? I would hold off on merging this/creating a new epics-in-docker release if we are likely to see further fixes being added.

PR #122, opened 2025-08-13
base: add linStat module.

ericonr:
This is a Linux-only alternative to iocStats that provides more information when possible.


ericonr:
Musl build should break due to mallinfo usage in linStat; will submit a patch for this.

PR #123, opened 2025-08-22
base: add lakeshore340 module

guirodrigueslima:
The lakeshore340 directory can be build as a standard EPICS support module.


gustavosr8:
Was making some tests and pushed to the wrong upstream, sorry

PR #127, opened 2025-10-23
base: add twincat-ads module

MaikTheWay:
Install twincat-ads separating modules and IOC

- Recursive clone of twincat-ads to include Beckhoff ADS sources (download/binary doesn't bring this submodule).
- Split twincat into two folders: ads-modules (modules) and ads-ioc (example IOC), and adapt configure/RELEASE (-include).


CHANGES.md
```diff
@@ -6,6 +6,8 @@
 
 * base: add lakeshore340 module. by @guirodrigueslima in
   https://github.com/cnpem/epics-in-docker/pull/123
+* base: add twincat-ads module. by @MaikTheWay in
+  https://github.com/cnpem/epics-in-docker/pull/126
```
gustavosr8: Sorry, I misdirect you here. This PR is #126

base/install_twincat_ads.sh
```diff
@@ -0,0 +1,39 @@
...
+
+cd ..
+
+install_module ads-modules ADS "
+EPICS_BASE
+ASYN
+"
+
+mkdir ads-ioc && cd ads-ioc
+yes ads | /opt/epics/base/bin/linux-x86_64/makeBaseApp.pl -i -t example
```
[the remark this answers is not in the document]
MaikTheWay: Changes finalized, new commit pushed.

base/install_twincat_ads.sh
```diff
@@ -0,0 +1,39 @@
...
+
+mv twincat-ads ads-modules
+cd ads-modules
+rm -rf .git*
+
+sed -i \
+    -e 's|^include|-include|' \
+    configure/RELEASE
+
+cd ..
```
gustavosr8: I'm not sure if you really need this. The `install__module` function already goes to the module directory and [overwrite `configure/RELEASE`](https://github.com/cnpem/epics-in-docker/blob/8f187706de34abc6e0f13e4a23395a1151c0ace0/base/install-functions.sh#L41).

gustavosr8 (changes requested):
Please squash 36dbfdad3c6e41d1bcc23076646b6b8db4e480c1 in the one adding those lines.

You can [rewrite your branch history](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) to accomplish this.

gustavosr8:
Github dropped some of my review comments :melting_face:. Will try to remember then again.

henriquesimoes:
As we discussed in depth in person, we'd like to move forward a more generic and reusable manner of dealing with IOCs inside epics-in-docker; especially in this case where the upstream doesn't provide a standard example IOC.

I'm marking this PR as draft until we can tackle such improvements (in another PR), which would be directly applicable for this module's IOC.

ericonr:
The changes were implemented correctly, but on the wrong commit. I will fix this up now, but you can use it as an example for other times :)

ericonr (approved):

PR #132, opened 2026-01-05
base: add motorSmarAct module.

guirodrigueslima:
Only the motorSmarAct module was compiled in the motor folder, because the iocs compilation structure will be changed in the future.


gustavosr8:
Depends on https://github.com/cnpem/epics-in-docker/pull/133 being merged

CHANGES.md
```diff
@@ -29,11 +29,9 @@ for that release.
   * This approach is experimental and intended for internal use within 
     the epics-in-docker project only. It is not yet considered stable 
     or supported for external use.
-* base: add motorSmarAct module. by @guirodrigueslima in
+* base: add motorSmarAct module and image. by @guirodrigueslima in
   https://github.com/cnpem/epics-in-docker/pull/132
 
-## v0.14.1
-
```
ericonr, line 139: This shouldn't be removed here.

images/docker-compose-motorsmaract.yml
```diff
@@ -0,0 +1,27 @@
+services:
+  motor-smaract:
+    image: ghcr.io/cnpem/motor-smaract-epics-ioc:$TAG
+    build:
+      context: ./
+      dockerfile: ../Dockerfile
+      target: dynamic-link
```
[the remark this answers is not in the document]
ericonr: If it's a new image, I would prefer we start with static. Assuming there are few deployments depending on how the image is laid out.

base/install_motor.sh
```diff
@@ -17,6 +17,8 @@ download_from_github epics-motor motorPIGCS2 $PIGCS2_VERSION
 download_from_github cnpem motorNewport $NEWPORT_VERSION
 download_from_github epics-motor motorParker $PARKER_VERSION
 
+download_from_github epics-motor motorSmarAct $SMARACT_VERSION
```
gustavosr8: Why is it apart from the other `download_from_github` calls? i.e.- why there is an empty line above?

PR #134, opened 2026-01-08
base: Add elcomat3000 module

guirodrigueslima:


ericonr (commented):
The first commit should mention "we are using the latest tag from the main branch since the repository does not have any tags"

gustavosr8:
> We are using the latest tag from the main branch since the repository does not have any tags.

Suggestion: 

We are using the last commit from the main branch since the repository does not have any tagged version, and it seems stable enough.

ericonr:
I changed the executable name to `elcomat3000` instead of `Elcomat3000`. Now I'm not sure which one is better, though.

gustavosr8 (approved):

PR #138, opened 2026-02-09
Update module motorPIGCS2

guirodrigueslima:
The motorPIGCS2 module was updated to a new tag, adding support for the E-727 and C-887 controllers. The pigcs2 IOC build was removed due to the new IOC build support (#133).


ericonr (commented):
Same comment about the commit message as #139

PR #141, opened 2026-03-04
base: add support for ADEiger module.

gustavosr8:
# Description

Add support for ADEiger module, in order to operate Pilatus4.

ZeroMQ library development package is required in the build stage of ADEiger module.

There are some unreleased features we would like to use, regarding Pilatus4 support, what makes necessary a checkout after initializing the submodule. The EDEIGER_VERSION points to the latest commit, but should point to R3-6 when the version were tagged.

# Checklist


ericonr (commented):
commit message:

> base: add support for ADEiger module.

base: add ADEiger module.

> ZeroMQ library development package is required in the build stage of
ADEiger module.

This module uses libzmq to interact with the detector data stream.

(this way you justify the presence of the -dev package and the runtime package in your runtime image)

> There are some unreleased features we would like to use, regarding
Pilatus4 support, what makes necessary a checkout after initializing the
submodule. The EDEIGER_VERSION points to the latest commit, but should
point to R3-6 when the version were tagged.

Since Pilatus4 support hasn't been released yet and we are interested in using
it, we are using the latest commit from the master branch instead of a tagged
release.

ericonr (commented):
I think tagging a new release is pending on further driver changes, right? So we have to use a commit for now.

ericonr (commented):
nit:

> [1]: https://github.com/areaDetector/ADEiger/pull/91

I prefer `[1] <link-content>` directly, without the colon

ericonr (approved):

PR #145, opened 2026-03-31
images: add softIOC image

guirodrigueslima:
Add docker-compose-softioc.yml for running a softIOC, including EPICS Base, sscan, calc, and pyDevSup modules.

### DIFF
--- a/.github/workflows/base-image.yml
+++ b/.github/workflows/base-image.yml
@@ -68,6 +68,7 @@ jobs:
             docker-compose-ca-gateway.yml
             docker-compose-pvagw.yml
             docker-compose-rgamv2.yml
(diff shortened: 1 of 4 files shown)


ericonr (commented):
commit message:

> Add docker-compose-softioc.yml for running a softIOC, including EPICS
Base, sscan, calc, and pyDevSup modules.

I can read this from the diff; what's interesting is for the commit to justify these choices. e.g.

> This IOC includes modules which are interesting for IOCs which interact with other IOCs through their PVs. sscan, calc and pyDevSup can be used to implement complex database logic for such cases.

ericonr:
I've fixed the commit to address my comments; please take a look at the final one later.

PR #151, opened 2026-04-08
Install templates by default in lnls-create-ioc

ericonr:
# Description

[base: always write Db/Makefile in lnls-create-ioc](https://github.com/cnpem/epics-in-docker/commit/70afdcbe93395ab8bd8ffdf78e91ecc26b1c7696)

This will be used to install some templates into the created IOCs by
default.

# Checklist

- [X] Describe your user-facing changes in `CHANGES.md`, following the format
- [X] Loading the records was tested with the rgamv2 image

### DIFF
--- a/base/lnls-create-ioc.sh
+++ b/base/lnls-create-ioc.sh
@@ -81,6 +81,12 @@ echo "TOP=../..
 include \${TOP}/configure/CONFIG
 " > "$MAKEFILE"
 
+echo "
+DB_INSTALLS += \${RECCASTER}/db/reccaster.db
+DB_INSTALLS += \${AUTOSAVE}/db/save_restoreStatus.db
+DB_INSTALLS += \${DEVIOCSTATS}/db/iocAdminSoft.db
+" >> "$MAKEFILE"
+
 if [ -n "${IOC_DBS:-}" ]; then
     for db in $IOC_DBS; do
         echo "DB_INSTALLS += $db" >> "$MAKEFILE"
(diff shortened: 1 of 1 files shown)


base/lnls-create-ioc.sh
```diff
@@ -90,6 +95,7 @@ fi
 echo "
 include \${TOP}/configure/RULES
 " >> "$MAKEFILE"
+
 # Copy command file(s) or directory(ies)
 if [ -n "${IOC_CMDS:-}" ]; then
```
henriquesimoes, line 101: This new line belongs to the first commit.

base/lnls-create-ioc.sh
```diff
@@ -81,6 +81,11 @@ echo "TOP=../..
 include \${TOP}/configure/CONFIG
 " > "$MAKEFILE"
 
+echo "
+DB_INSTALLS += \${AUTOSAVE}/db/save_restoreStatus.db
+DB_INSTALLS += \${DEVIOCSTATS}/db/iocAdminSoft.db
```
henriquesimoes, line 87: Should we be installing `$(RECCASTER)/db/reccaster.db` or `$(PVXS)/db/softIocPVXExit.db` as well? It isn't clear to me what criteria you are using to call a database file "relevant".

ericonr: I didn't know about either of those; `reccaster.db` should be included, it can provide reccaster status. `softIocPVXExit.db` is used by https://github.com/epics-base/pvxs/blob/1d481d5be1db1e1bcc934c82a8bcec162c3ac31f/qsrv/softMain.cpp#L90 , so it's not relevant for us. I will update the commit accordingly.

PR #157, opened 2026-04-24
Add motorZaberMotion module and image

guirodrigueslima:
# Checklist


base/Dockerfile
```diff
@@ -13,6 +13,7 @@ RUN echo 'APT::Sandbox::User "root";' > /etc/apt/apt.conf.d/90-disable-sandbox.c
 RUN apt update -y && \
     apt install -y --no-install-recommends \
         build-essential \
+        curl \
```
gustavosr8: Missing some explanation in the commit message of **why** it was necessary to include `curl`.

images/docker-compose-motorzaber.yml
```diff
@@ -0,0 +1,23 @@
...
+      context: ./
+      dockerfile: ../Dockerfile
+      target: dynamic-link
+      labels:
+        org.opencontainers.image.source: https://github.com/cnpem/epics-in-docker
+      args:
+        IOC_CREATE: 1
+        REPONAME: ZaberMotion
+        RUNDIR: /opt/ZaberMotion/iocBoot/iocZaberMotion
+        APP_DIRS: /opt/epics/modules/motorZaberMotion
```
gustavosr8: > Due to an issue with prune [...]

Could you please open an issue with more detailed information about this?

gustavosr8:
As you are adding an image, please use the correspondent checklist in the description.

Also, since you choose to install the module with a dedicated shell script, don't think you need to first use the `install_from_github` method and then change it. The PR should be atomic.

base/install_zaber.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+set -ex
+
+. $EPICS_IN_DOCKER/install-functions.sh
+. $EPICS_IN_DOCKER/zaber_versions.sh
+
+ZML_DIR=/opt/epics/modules/motorZaberMotion/zaberMotionSupport
```
gustavosr8: `/opt/epics/modules` could be `$EPICS_MODULES_PATH`

base/install_zaber.sh
```diff
@@ -0,0 +1,27 @@
...
+ZML_ZIP=${ZML_DIR}/ZaberMotionCppSupport-${ZABER_MOTION_LIBRARY_VERSION}.zip
+ZML_URL="https://www.zaber.com/support/software-downloads.php?product=zml_cpp_support&version=${ZABER_MOTION_LIBRARY_VERSION}"
+
+download_from_github zabertech motorZaberMotion $ZABER_MOTION_VERSION
+
+wget -O "${ZML_ZIP}" "${ZML_URL}"
```
gustavosr8, line 14: I think that the commit could mention that the upstream module uses curl, but we choose to use wget since it was already installed

base/install_zaber.sh
```diff
@@ -0,0 +1,27 @@
...
+unzip -q -o "${ZML_ZIP}" -d "${ZML_DIR}"
+rm -f "${ZML_ZIP}"
```
ericonr, line 16: Can't you use `lnls-get-n-unpack` here?

base/install_zaber.sh
```diff
@@ -0,0 +1,27 @@
...
+
+cp -a ${ZML_DIR}/ZaberMotionCppSupport/lib/linux_x64/* /usr/local/lib/
+cp -r ${ZML_DIR}/ZaberMotionCppSupport/include/zaber/motion/* /usr/local/include/
+ldconfig
```
ericonr: Are you sure `ldconfig` was necessary? It should only make a difference in the final runtime image

base/install_zaber.sh
```diff
@@ -0,0 +1,27 @@
...
+
+sed -i '/zaberMotionSupport/s/^/# /' motorZaberMotion/Makefile
```
ericonr, line 21: I think it's worth adding a comment explaining this

Why comment instead of simply deleting the lines?

base/motor_versions.sh
```diff
@@ -3,3 +3,4 @@ NEWPORT_VERSION=R1-3-lnls
 PIGCS2_VERSION=R1-3
 PMAC_VERSION=2-7-9
 PARKER_VERSION=R1-2-1
+ZABER_MOTION_VERSION=v2.0.0
```
ericonr: it's still here

images/docker-compose-motorzaber.yml
```diff
@@ -0,0 +1,22 @@
+services:
+  motor-zaber:
+    image: ghcr.io/cnpem/motor-zaber-epics-ioc:$TAG
+    build:
+      context: ./
+      dockerfile: ../Dockerfile
+      target: dynamic-link
```
ericonr, line 7: Can you pull all the necessary dbs into this image? Are there too many? I would like to eventually look into making it `static-link`

base/install_zaber.sh
```diff
@@ -0,0 +1,27 @@
...
+ZML_URL="https://www.zaber.com/support/software-downloads.php?product=zml_cpp_support&version=${ZABER_MOTION_LIBRARY_VERSION}"
+
+download_from_github zabertech motorZaberMotion $ZABER_MOTION_VERSION
+
+wget -O "${ZML_ZIP}" "${ZML_URL}"
+unzip -q -o "${ZML_ZIP}" -d "${ZML_DIR}"
+rm -f "${ZML_ZIP}"
+
+cp -a ${ZML_DIR}/ZaberMotionCppSupport/lib/linux_x64/* /usr/local/lib/
+cp -r ${ZML_DIR}/ZaberMotionCppSupport/include/zaber/motion/* /usr/local/include/
```
ericonr, line 19: Why `-a` and then `-r`? I don't think there were symlinks in lib, were there?

base/install_zaber.sh
ericonr: The tool used to download is not the most relevant part. The main discussion is that their Makefile downloads a tarball by itself, out of our control. We'd like to avoid that.

base/install_zaber.sh
guirodrigueslima: I don't know, I'll try it and then make changes if necessary.

base/install_zaber.sh
guirodrigueslima: I prefer to leave the comment to keep the information within the file. What do you think?

base/install_zaber.sh
guirodrigueslima: Because we have symlinks on lib and we have subfolder on includes.

base/install_zaber.sh
guirodrigueslima: I talked to @henriquesimoes about this the other day, I'm going to change it.

CHANGES.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### New features
+
```
gustavosr8, line 6: Already have `New features`

CHANGES.md
```diff
@@ -2,6 +2,11 @@
...
+* base: add motor zaber module. by @guirodrigueslima in
```
gustavosr8, line 7:
``` suggestion
* base: add motorZaberMotion module. by @guirodrigueslima in
```

base/Dockerfile
```diff
@@ -73,6 +73,13 @@ RUN $EPICS_IN_DOCKER/install_area_detector.sh
 COPY motor_versions.sh install_motor.sh $EPICS_IN_DOCKER
 RUN $EPICS_IN_DOCKER/install_motor.sh
 
+ARG DEBIAN_VERSION
+COPY opcua_versions.sh install_opcua.sh $EPICS_IN_DOCKER
+RUN $EPICS_IN_DOCKER/install_opcua.sh
+
```
gustavosr8, line 79: misleading rebase?

base/install_zaber.sh
```diff
@@ -0,0 +1,26 @@
...
+ZML_DIR=${EPICS_MODULES_PATH}/motorZaberMotion/zaberMotionSupport
+ZML_ZIP=${ZML_DIR}/ZaberMotionCppSupport-${ZABER_MOTION_LIBRARY_VERSION}.zip
+ZML_URL="https://www.zaber.com/support/software-downloads.php?product=zml_cpp_support&version=${ZABER_MOTION_LIBRARY_VERSION}"
+
+download_from_github zabertech motorZaberMotion $ZABER_MOTION_VERSION $ZABER_MOTION_SHA256
+
+wget -O "${ZML_ZIP}" "${ZML_URL}"
+unzip -q -o "${ZML_ZIP}" -d "${ZML_DIR}"
+rm -f "${ZML_ZIP}"
+
```
gustavosr8, line 17: Isn't `lnls-get-n-unpack` better here?

gustavosr8 (commented):
I think that the first and the last commit should be squashed. Don't see a reason for them to be separated.

base/install_zaber.sh
ericonr: The issue is get-n-unpack uses the file extension instead of querying the contents to discover what compression engine to use.

We would need to move to `bsdtar` or use `file` here in order to fix this (I'd prefer the bsdtar solution).

base/zaber_versions.sh
```diff
@@ -0,0 +1,4 @@
+ZABER_MOTION_LIBRARY_VERSION=9.0.0
+
+ZABER_MOTION_VERSION=v2.0.0
+ZABER_MOTION_SHA256=f714419345d0917ddc43e2af3a8e4cda3df0c78f4fbbee7d41ebd8335caa0ec3
```
ericonr, line 4: The library should have checksum verification as well. I think our download wrapper needs to be improved to support this use case, instead of having it implemented manually.

ericonr (changes requested):

PR #165, opened 2026-05-25
Complex updates for May 2026 (OPCUA, epics-base, Debian 13, Alpine 3.23)

### LINKED ISSUE #23: Use newer version of procServ?
Using a newer version of procServ would mean having support for the `--once/oneshot` flag (idk which one they settled on), which would be useful to allow tighter control over container and IOC lifetime.

It would require building procServ ourselves, but that should be pretty simple.

At the same time, most orchestration stuff wouldn't be able to rely on this, because they can't assume the container was built using this repo (and uses a recent enough version of this repo).

### LINKED ISSUE #109: Build the OPCUA module using open62541
This will allow us to build any version of the OPCUA module with any version of epics-base and the underlying distro. Having to move everything in lockstep adds more workload and can block progress.

Will require coordinated testing with other groups to ensure it's working correctly.

ericonr:
# Checklist

- [X] Describe your user-facing changes in `CHANGES.md`, following the format

~@henriquesimoes as our resident Debian expert, I'd appreciate some official guidance on whether we should use the `t64` suffix on package names or not.~


CHANGES.md
```diff
@@ -23,6 +23,10 @@
 
 ### New features
 
+* base: update opcua to 0.11.2, use open62541 by @ericonr in
```
gustavosr8: Just noticed we were hardcoding the `v` in `base/install_opcua.sh`, so nit: `0.11.2` -> `v0.11.2`

base/install_modules.sh
```diff
@@ -149,3 +149,24 @@ install_module twincat-ads ADS "
...
+make -j$JOBS
+make install
+cd ../../
+rm -rf open62541
+
+download_from_github epics-modules opcua $OPCUA_VERSION
+cat << EOF > opcua/configure/CONFIG_SITE.local
+OPEN62541=/usr/local/lib
+OPEN62541_DEPLOY_MODE=SYSTEM
+OPEN62541_USE_CRYPTO=YES
```
gustavosr8, line 166: Don't we need to install `libcrypto`?

gustavosr8 (commented):
Since we are touching pyDevSup, wdyt about updating the module org, since it now lives under `epics-modules` instead of `mdavidsaver`?

CHANGES.md
ericonr: I think I'm against hardcoding the `v` in install scripts. Will make it consistent across everything

base/install_modules.sh
ericonr: It's contained in `libssl-dev`!

CHANGES.md
ericonr: Ah, I had made it consistent. Only the changelog was inconsistent.

gustavosr8 (approved):

CHANGES.md
```diff
@@ -34,6 +34,8 @@
 
 ### New features
 
+* base: update to Debian 13 by @ericonr in
+  https://github.com/cnpem/epics-in-docker/pull/165
```
henriquesimoes, line 38: This upgrade gives us procServ 2.8.0, which means #23 is satisfied. Maybe it is worth adding a subitem about this.

I just linked this PR as a resolution to the mentioned issue.

henriquesimoes (approved):

henriquesimoes (approved):

PR #166, opened 2026-06-03
base: enable build with Boost.

gustavosr8:
The areaDetector plugin-test utility uses the Boost Unit Testing Framework. To correctly build the application, we need to install the lib and enable its setting `WITH_BOOST` and `BOOST_EXTERNAL`

# Checklist

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -16,6 +16,8 @@ RUN apt update -y && \
         cmake \
         git \
         libaravis-dev \
+        libboost-test-dev \
+        libboost-system-dev \
         libevent-dev \
         libnet-dev \
         libpcap-dev \
--- a/base/install_area_detector.sh
+++ b/base/install_area_detector.sh
@@ -53,7 +53,8 @@ ln -s RELEASE.local RELEASE_LIBS.local
 echo "
 BUILD_IOCS=NO
 
-WITH_BOOST=NO
+WITH_BOOST=YES
+BOOST_EXTERNAL=YES
 
 WITH_PVA=YES
 WITH_QSRV=YES
@@ -106,6 +107,8 @@ cd -
 
 make -j${JOBS}
 make clean
+
+ADCore/bin/*/plugin-test
 
 cd ..
 


CHANGES.md
```diff
@@ -44,6 +44,8 @@
   https://github.com/cnpem/epics-in-docker/pull/164
   * We were on version 1.3.0. Changelog is available in
     https://epics-base.github.io/pvxs/releasenotes.html
+* base: enable areaDetector build with Boost by @gustavosr8 in
+  https://github.com/cnpem/epics-in-docker/pull/166
```
ericonr: This shouldn't be relevant to end-users.

ericonr (commented):
I don't think the commit properly motivates the change. What's your goal for building `plugin-test`?

Nit: enable its usage with ...

And there's a missing period in the commit body.

CHANGES.md
gustavosr8: I think it is relevant to mention since now the built image has the `plugin-test` utility. One desiring to use this feature may find it relevant, IMO.

Can be more explicit about this in the notes, though.

ericonr: I don't agree at all. This utility is for testing the ADCore project, and it is therefore relevant to developers of said project (so it runs in CI for them) and for integrators of that project (who need to build it and run it in their environment).

We, as integrators, can (and maybe should?) add a step to the areaDetector build which also runs the tests; that's just an improvement to our internal processes to ensure our build is working correctly. That's where the usefulness ends.

Using the resulting build image as a stepping stone for testing changes to ADCore, and having tests already integrated with that, is not a documented or supported use-case (which may change; I have found epics-in-docker really useful for experimentation), and though it may be used as a justification in the commit message (since it's useful to us, as integrators and developers), it isn't part of the interface we provide to external users.

ericonr (changes requested):
> When integrating a plugin in an areaDetector application, it can be
useful to ensure it is properly working as expected. Since the lib
already give us an application that run some tests, we can just launch
it.

I don't think this exactly explains our intentions. I would go simpler: "the project provides tests for some plugins, which we can use to ensure our build configuration hasn't affected the tested functionality"

I would also make this PR a single commit: "base: run ADCore tests", where plugin-test is the only (currently) available test, and adding boost/enabling boost is just a requirement for the tests to be available.

ericonr (approved):

PR #168, opened 2026-06-09
Add checksum verification for all modules

gustavosr8:
Continuation of #80


henriquesimoes:
Can you please deal with [this pending documentation point](https://github.com/cnpem/epics-in-docker/pull/80#pullrequestreview-3384826022) I raised in the other thread?

base/lnls-get-n-unpack.sh
```diff
@@ -31,19 +33,43 @@ parse_arguments() {
...
     echo "$dest $@"
 }
 
+shacheck() {
+    downloaded_file=$1
+    sha=$2
+    url=$3
+
+    if ! echo $sha $downloaded_file | sha256sum -c; then
+        echo "ERROR: SHA $sha for URL $url does not match."
```
henriquesimoes:
```suggestion
        computed=$(sha256sum $downloaded_file | cut -d ' ' -f 1)
        echo >&2 "error: checksum mismatch for '$url': expected '$sha', computed '$computed'."
```

PS: I'm not sure how the source should be wrapped.

base/lnls-get-n-unpack.sh
```diff
@@ -31,19 +33,43 @@ parse_arguments() {
 
     # Throw extraction mode argument away
     shift
+
+    # Check if we have an even number of arguments (N * (url + sha256sum))
+    if [ $(($# % 2)) -ne 0 ]; then
+        echo >&2 "ERROR: Odd number of arguments detected. Something is wrong."
```
henriquesimoes:
```suggestion
        echo >&2 "error: mismatching number of URL and hash arguments."
```

henriquesimoes (commented):
Just two suggestions regarding the error messages. Other than that, it looks fine by me.

base/lnls-get-n-unpack.sh
ericonr: No wrapping, we can live without it for error messages

base/lnls-get-n-unpack.sh
ericonr: Thanks!

I did keep the capitalized ERROR, though.

Without color output, I think it's needed.

ericonr (approved):

PR #169, opened 2026-06-10
add industry pack module and image ioc

guirodrigueslima:
# Checklist

## If adding a new IOC image


images/docker-compose-industry-pack.yml
ericonr, line 1: Do you think it's worth defining an image only for stanfordsr570 when we can instead have an image for all devices supported by `ip`?

guirodrigueslima: I believe we can develop a single IOC that encompasses all IP projects. However, I intend to keep the target as dynamic, since there will be many files to add

base/modules_versions.sh
```diff
@@ -27,3 +27,4 @@ TWINCAT_ADS_VERSION=v2.1.3
 ADS_VERSION=v22
 OPEN62541_VERSION=v1.3.17
 OPCUA_VERSION=v0.11.2
+IP_VERSION=R2-22
```
gustavosr8, line 85: I've noticed that this version is more than 3 years old. Any of the subsequent commits are wanted?

CHANGES.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* base: add ip module and image. by @guirodrigueslima in
+  https://github.com/cnpem/epics-in-docker/pull/169
+
```
gustavosr8: I think that this should live under the `New features` header, no?

PR #170, opened 2026-06-18
base: add SeaBreeze library and Ocean Optics support module

GuPecz:
Add support for SeaBreeze library, in order to operate QEPro spectrometer.

SeaBreeze depends on libusb-0.1.

# Checklist


base/install_modules.sh
```diff
@@ -171,3 +171,9 @@ rm -rf opcua/exampleTop
 install_module opcua OPCUA "
 EPICS_BASE
 "
+
+download_from_github jwlodek SeaBreeze $SEABREEZE_VERSION $SEABREEZE_SHA256
+cd SeaBreeze
+make -j$JOBS
+cd ..
```
henriquesimoes:
```suggestion
make -C SeaBreeze -j$JOBS
```

base/install_modules.sh
```diff
@@ -171,3 +171,9 @@ rm -rf opcua/exampleTop
...
+rm -rf seabreeze
```
henriquesimoes: `SeaBreeze` (camel case)?

Does it install everything in the system in its default Make target? If so, can you please document that in the commit message?

CHANGES.md
```diff
@@ -34,6 +34,7 @@
 
 ### New features
 
+* base: add SeaBreeze library by @GuPecz in https://github.com/cnpem/epics-in-docker/pull/170
```
henriquesimoes: Hard-warp in 80 columns, please.

henriquesimoes (commented):
What are the plans regarding the EPICS support module? Will it be included in epics-in-docker as well or will it be built externally?

ericonr:
> What are the plans regarding the EPICS support module? Will it be included in epics-in-docker as well or will it be built externally?

The plan is to include it in epics-in-docker. IMO, we should only merge this along with the IOC.

ericonr:
I would consider making it a single commit. We only want seabreeze if we have the OceanOptics driver.

GuPecz:
We are settling to port the Ocean Optics support module to use the OceanDirect SDK instead of SeaBreeze. Until the port is done, we'll still be using the module's original implementation with SeaBreeze, so the PR will be marked as draft.

PR #171, opened 2026-06-24
base: run plugin-test from /tmp.

gustavosr8:
The plugin-test utility generates some xml and HDF5 files when performing the tests. When executed from areaDetector root directory, those files weren't reached by lnls-prune-artifacts and increases the final image in ~6MB. Run the tests from /tmp so the temporary artifacts are excluded when became useless.

# Checklist

### DIFF
--- a/base/install_area_detector.sh
+++ b/base/install_area_detector.sh
@@ -108,7 +108,11 @@ cd -
 make -j${JOBS}
 make clean
 
-ADCore/bin/*/plugin-test
+(
+    cd $(mktemp -d)
+    ${EPICS_MODULES_PATH}/areaDetector/ADCore/bin/*/plugin-test
+    rm -rf $PWD
+)
 
 cd ..
 


ericonr (changes requested):
> base: run plugin-test from /tmp.

base: run areaDetector plugin-test from /tmp

> increases the
final image in ~6MB. Run the tests from /tmp so the temporary artifacts
are excluded when became useless.

increased the final image size by ~6MB. If we run the tests and dump the
artifacts in /tmp, they won't be copied to the IOC images.

(if we actually want to not ship these files at all, you're missing an `rm` step)

base/install_area_detector.sh
```diff
@@ -108,7 +108,11 @@ cd -
 make -j${JOBS}
 make clean
 
-ADCore/bin/*/plugin-test
+(
+    cd /tmp
+    ${EPICS_MODULES_PATH}/areaDetector/ADCore/bin/*/plugin-test
+    rm -rf *
```
ericonr:
```
    cd $(mktemp -d)
    ${EPICS_MODULES_PATH}/areaDetector/ADCore/bin/*/plugin-test
    rm -rf $PWD
```

ericonr (changes requested):
> Run the tests from /tmp and remove everything
stored on this directory.

Run the tests in a temporary directory and clean it up afterwards.

ericonr (approved):

PR #173, opened 2026-06-26
Include Xspress3 module and IOC

guirodrigueslima:
# Checklist

- [x] Follow the commit format specified in `CONTRIBUTING.md`;
- [x] Describe your user-facing changes in `CHANGES.md`, following the format
      established by previous entries.

## If adding a new IOC image

- [x] Create a compose file for the new IOC under the `images` directory;
- [x] List the new compose file in `.github/workflows/base-image.yml` under the
      `files` key in the `Build and push included IOC images` step;
- [x] Add an entry in `README.md`, under `Included IOC images`, with the IOC
      name and its fully qualified container image name.

PR #174, opened 2026-06-29
base: prune broken symbolic links correctly

henriquesimoes:
Shared libraries may be installed with symbolic links for their version components, so that build systems can pull the library based on their requirements on the API. During the prune phase, lnls-prune-artifacts will get rid of the main shared library as well as its corresponding links whenever the library isn't used, so the filesystem is consistent afterward.

That works as expected whenever the user application doesn't link to the library at all or links to the library using a proper symbolic link for *.so file. When the project ships the unversioned library as a hard copy, and such library is used by an application, an unwanted scenario is reached: remove_unused_shared_libraries() will decide to remove the *.so copy (because it is an ELF binary, but isn't used at runtime, as only its versioned copy is) and, thus, it will remove it along with its versioned friend; although the versioned one is to be kept.

Instead of relying on this fragile heuristics of removing every library with its corresponding link, which depends on the project build system to install symbolic links for versioned libraries, only remove the files checked against the "keep_paths". This ensures we won't ever remove a shared library in use. Still, to keep the filesystem consistent, introduce another short pruning step, which targets at finding only leftover broken links.

Broken links are queried using a GNU-specific flag (-xtype) for find(1), so that we check the existence of the linked file, not the link itself [1]. This is not a problem because pruning is relevant only for Debian-based build images, in which GNU tooling is available. Placing the documentation comment close to the find(1) would be possible with a here-string, but it would introduce a final newline [2], which would require us to treat empty values in the loop. As it is a short loop body, just use a process substitution instead.

Also, loop with a while-read pattern, so that we properly deal with files with spaces in their path.

[1] https://unix.stackexchange.com/a/38691
[2] https://www.gnu.org/software/bash/manual/html_node/Redirections.html#Here-Strings

Fixes: 22ac3c6 (ioc: prune shared libraries from non-static builds., 2024-04-11)

---

This should fix the build issue affecting #173.

# Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,11 @@
 
 ### Bug fixes
 
+* base: prune broken symbolic links correctly in
+  https://github.com/cnpem/epics-in-docker/pull/174
+  * Improve the shared library symbolic link removal to prevent the mistaken
+    pruning of linked libraries when there are hard-copies of shared objects
+    instead of proper symlinks.
 * Rectify pyDevSup IOC dependencies by @ericonr in
   https://github.com/cnpem/epics-in-docker/pull/165
   * Refer to the README for updated guidance on pyDevSup IOCs.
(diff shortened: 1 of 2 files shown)


ericonr:
Will be merged in #173

PR #183, opened 2026-07-29
images: add modbus image

guirodrigueslima:
# Checklist

## If adding a new IOC image

### DIFF
--- a/.github/workflows/base-image.yml
+++ b/.github/workflows/base-image.yml
@@ -63,6 +63,7 @@ jobs:
             docker-compose-epics-base.yml
             docker-compose-industry-pack.yml
             docker-compose-opcua.yml
+            docker-compose-modbus.yml
             docker-compose-motorgalil.yml
             docker-compose-motornewport.yml
             docker-compose-motorpigcs2.yml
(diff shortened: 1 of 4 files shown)


ericonr:
Who's the intended user for this?

guirodrigueslima:
I recently encountered a modbus application that uses only binaries via NFS with st.cmd and database files in the configuration repository. Therefore, I would like to add a standard modbus binary to this dedicated system.

ericonr:
Is it a situation similar to what we have with OPCUA IOCs, where the controlled "device" is specific to the application? Or should it become a dedicated module?

guirodrigueslima:
> Is it a situation similar to what we have with OPCUA IOCs, where the controlled "device" is specific to the application? Or should it become a dedicated module?

Exactly, like OPCUA and StreamDevice images. And we’ll also have an image for running modbus tests. 

A new module would be dedicated to new library and/or file dependencies. Everything we need is already in epics-in-docker.

gustavosr8 (approved):

ericonr:
> StreamDevice images

We don't have StreamDevice images because they are specific to devices, not installations.

Please show me your intended use case offline.

ericonr:
As discussed offline, this won't be merged. modbus support should be included in support modules.

PR #184, opened 2026-08-13
base: change IS_IOC_AREADETECTOR flag

guirodrigueslima:
Changes the IS_IOC_AREADETECTOR flag, used when building AreaDetector IOCs, to accept numeric values (0 or 1) instead of strings (false or true).

This change is intended to standardize the project, since the other binary flags also use numeric values.

# Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,11 @@
   https://github.com/cnpem/epics-in-docker/pull/186
   * Fix IOC crash caused by connection problems.
 
+### Breaking changes
+
+* base: take only numeric values for IS_IOC_AREADETECTOR. by @guirodrigueslima in
+  https://github.com/cnpem/epics-in-docker/pull/184
+
 ## v0.16.0
 
 This update is recommended for most users. It greatly improves iocsh
--- a/base/lnls-create-ioc.sh
+++ b/base/lnls-create-ioc.sh
@@ -69,7 +69,7 @@ echo "
 \${PROD_NAME}_SRCS_DEFAULT += ${IOC_NAME%App}Main.cpp
 " >> "$MAKEFILE"
 
-if [ "${IS_IOC_AREADETECTOR:-false}" = "true" ]; then
+if [ "${IS_IOC_AREADETECTOR:-0}" -eq 1 ]; then
     echo "include \${ADCORE}/ADApp/commonDriverMakefile" >> "$MAKEFILE"
 fi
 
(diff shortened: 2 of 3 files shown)


CHANGES.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### New features
+
```
gustavosr8: More than a new feature, I think that this is actually a Breaking change, isn't it?

gustavosr8:
Commit suggestion:

> base: change IS_IOC_AREADETECTOR flag to accept numeric values.

henriquesimoes:
> Commit suggestion:
> 
> > base: change IS_IOC_AREADETECTOR flag to accept numeric values.

I agree it deserves a more descriptive message. However, we aren't moving to "accept" numeric values. We are making this the only way of specifying the values. How about the following?

`base: take only numeric values for IS_IOC_AREADETECTOR`

gustavosr8:
> I agree it deserves a more descriptive message. However, we aren't moving to "accept" numeric values. We are making this the only way of specifying the values. How about the following?
> 
> `base: take only numeric values for IS_IOC_AREADETECTOR`

Agreed!

CHANGES.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Breaking changes
+
+* base: change IS_IOC_AREADETECTOR flag. by @guirodrigueslima in
+  https://github.com/cnpem/epics-in-docker/pull/184
+
```
ericonr, line 15: Is this breaking if we haven't officially released the new build method to external users?

guirodrigueslima: > Isso seria considerado uma quebra de compatibilidade se ainda não tivéssemos lançado oficialmente o novo método de compilação para usuários externos?

I wondered about that too, but we don't know who might already be using the project

guirodrigueslima: > Is this breaking if we haven't officially released the new build method to external users?

I wondered about that too, but we don't know who might already be using the project

gustavosr8: Justifying my suggestion, we documented the new IOC building method as a breaking change in the [v0.15.0 release notes](https://github.com/cnpem/epics-in-docker/blob/ad79d08d4d69520f9e3b95b9b6ef3a07bdfed006/CHANGES.md?plain=1#L122-L128), which introduced the `IS_IOC_AREADETECTOR` flag. Since we are changing this interface, this looks also a breaking change.

ericonr: @gustavosr8 the breaking change is that we changed the paths inside of the IOC images, not that the new build method was added. The actual addition of the build method was a new feature, which was explicitly called out as experimental and only for internal use.

https://github.com/cnpem/epics-in-docker/blob/ad79d08d4d69520f9e3b95b9b6ef3a07bdfed006/CHANGES.md?plain=1#L134-L138

ericonr:
The commit isn't about `base`, it's about `ioc`.

ericonr:
Also link to <https://github.com/containers/podman-compose/issues/1022> as one of the issues caused by using those specific strings.

PR #185, opened 2026-08-14
base: change symetrie path name

guirodrigueslima:
The module path name was changed because the substitution file uses the name SYMETRIE by default.

https://github.com/epics-motor/motorSymetrie/blob/main/SYMETRIE_EPICSApp/Db/SymetriePmac.substitutions

# Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,9 @@
   https://github.com/cnpem/epics-in-docker/pull/191
   *  python3-numpy doesn't install libpython3.13 as expected. Install it
      manually.
+* base: fix symetrie path variable. by @guirodrigueslima in
+  https://github.com/cnpem/epics-in-docker/pull/185
+  * This meant the provided substitutions file wasn't working correctly.
 
 ## v0.16.0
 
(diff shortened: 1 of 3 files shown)


gustavosr8:
Looks good to me, just some minor details

About the commit message:
- Missing the reference you pointed in the PR description in the commit message;
- `s/substitution/substitutions`.

About the Changelog Documentation:
- Why is it a breaking change instead of a bugfix?

gustavosr8:
I think that it is better to add the permalink as the reference:

https://github.com/epics-motor/motorSymetrie/blob/6233721f2a11d841db3cd0c493a6051228caa32b/SYMETRIE_EPICSApp/Db/SymetriePmac.substitutions

Besides this, what do you think about `s/change/fix` in the commit title? Does it works without this modification?

guirodrigueslima:
> I think that it is better to add the permalink as the reference:
> 
> https://github.com/epics-motor/motorSymetrie/blob/6233721f2a11d841db3cd0c493a6051228caa32b/SYMETRIE_EPICSApp/Db/SymetriePmac.substitutions
> 
> Besides this, what do you think about `s/change/fix` in the commit title? Does it works without this modification?

The branch name is fix_symetrie :smile:

gustavosr8 (approved):

PR #187, opened 2026-08-18
Correctly set the CLI tools for non 'end-user' usage

gustavosr8:
By default, debconf(1) frontend is defined as 'dialog', which is incompatible with containerization interactions, leading to a series of failures and fallbacks to different frontend modes till it reaches 'noninteractive'.

This can be avoided by defining `DEBIAN_FRONTEND=noninteractive`.

Also, as defined in the CLI man page, `apt` is designed as an end-user tool, and scripts should use `apt-get` instead.

In practice, get rid of the warnings:

```
debconf: unable to initialize frontend: Dialog
debconf: (TERM is not set, so the dialog frontend is not usable.)
debconf: falling back to frontend: Readline
debconf: unable to initialize frontend: Readline
debconf: (Can't locate Term/ReadLine.pm in @INC (you may need to install the Term::ReadLine module) (@INC entries checked: /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.40.1 /usr/local/share/perl/5.40.1 /usr/lib/x86_64-linux-gnu/perl5/5.40 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl-base /usr/lib/x86_64-linux-gnu/perl/5.40 /usr/share/perl/5.40 /usr/local/lib/site_perl) at /usr/share/perl5/Debconf/FrontEnd/Readline.pm line 8, <STDIN> line 16.)
debconf: falling back to frontend: Teletype
debconf: unable to initialize frontend: Teletype
debconf: (This frontend requires a controlling tty.)
debconf: falling back to frontend: Noninteractive
```

and

`WARNING: apt does not have a stable CLI interface. Use with caution in scripts.`

# Checklist

- [x] Follow the commit format specified in `CONTRIBUTING.md`

### DIFF
--- a/.github/workflows/base-image.yml
+++ b/.github/workflows/base-image.yml
@@ -82,6 +82,6 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - name: Install packages
-        run: sudo apt update -y && sudo apt install -y pre-commit shfmt
+        run: sudo apt-get update -y && sudo apt-get install -y pre-commit shfmt
       - name: Check pre-commit hooks
         run: pre-commit run --show-diff-on-failure --color=always --all-files
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,7 @@ ARG RUNTIME_PIP_PACKAGES
 
 ARG PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 ENV PATH=$PATH
+ENV DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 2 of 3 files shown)


gustavosr8:
For reference, motivated by the workflow logs starting [here](https://github.com/cnpem/cryo-load-leveler-epics-softioc/actions/runs/32151558893/job/95758432347#step:5:518).

GuPecz (approved):

ericonr (commented):
Call out in first commit that we already use it in the base image Dockerfile

PR #188, opened 2026-08-18
Module support

guirodrigueslima:
Hello everyone!

Attached is the implementation of the new method for building modules with epics-in-docker.

With this implementation, it will be possible to build a module from an epics-in-docker clone within the project itself and then create an IOC using that module.

To enable this behavior, simply set the IS_MODULE variable in the YAML file.

Below is an example of the YAML configuration I am using to build the ADGStreamer module:

```
services:
  ioc:
    image: ghcr.io/cnpem/adgstreamer
    build:
      context: ./
      dockerfile: docker/Dockerfile
      target: dynamic-link
      labels:
        org.opencontainers.image.source: https://github.com/cnpem/ADGStreamer
      args:
        BUILD_PACKAGES: libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev
        RUNTIME_PACKAGES: libxml2 libtiff6 gstreamer1.0-tools gstreamer1.0-plugins-base gstreamer1.0-plugins-good gstreamer1.0-plugins-bad gstreamer1.0-plugins-ugly gstreamer1.0-libav
        IS_MODULE: 1
        MODULE_NAME: ADGST
        REPONAME: ADGStreamer
        RUNDIR: /opt/ADGStreamer/iocBoot/iocADGStreamer
        IS_IOC_AREADETECTOR: 1
        IOC_DBDS:
          asyn.dbd
          ADGStreamer.dbd
        IOC_LIBS:
          ADGStreamer
          asyn
        IOC_DBS:
          $(ADGST)/db/ADGStreamer.template
```

I would appreciate your help in defining a documentation structure for this new feature in the README.

With this change, epics-in-docker will support both IOC builds and module builds directly from an epics-in-docker clone within the project.

# Checklist

### DIFF
--- a/base/lnls-build-module.sh
+++ b/base/lnls-build-module.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+set -ex
+
+. $EPICS_IN_DOCKER/install-functions.sh
+
+mkdir -p ${EPICS_MODULES_PATH}/${REPONAME}
+mv * ${EPICS_MODULES_PATH}/${REPONAME}
+install_module ${EPICS_MODULES_PATH}/${REPONAME} ${MODULE_NAME} "*"
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,8 @@
 
 * base: get all module paths. by @guirodrigueslima in
   https://github.com/cnpem/epics-in-docker/pull/188
+* base: add support for module builds. by @guirodrigueslima in
+  https://github.com/cnpem/epics-in-docker/pull/188
 
 ## v0.16.0
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,11 @@
 * base: update twincat-ads to v2.2.1 by @ericonr in
   https://github.com/cnpem/epics-in-docker/pull/186
   * Fix IOC crash caused by connection problems.
+  
+### New features
+
+* base: get all module paths. by @guirodrigueslima in
+  https://github.com/cnpem/epics-in-docker/pull/188
 
 ## v0.16.0
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -91,6 +91,8 @@ WORKDIR /opt/${REPONAME}
 
 COPY . .
 
+ARG IS_MODULE
+ARG MODULE_NAME
 ARG IOC_CREATE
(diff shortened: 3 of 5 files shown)


base/lnls-build-module.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+set -ex
+
+. $EPICS_IN_DOCKER/install-functions.sh
+
+mkdir -p ${EPICS_MODULES_PATH}/${REPONAME}
+mv * ${EPICS_MODULES_PATH}/${REPONAME}
+install_module ${EPICS_MODULES_PATH}/${REPONAME} ${MODULENAME} "*"
```
gustavosr8: I think that the ARG names here are a little misleading. In `install_module` we call `module_name` what here we are calling `REPONAME`, and here we call `MODULENAME` what is `dependency_name` in the script

guirodrigueslima: I was unsure about how to name a variable. What do you think of MODULEPATH? This must be the name of the module path.

guirodrigueslima: or MODULE_PATH_NAME ?

gustavosr8: Why not `MODULE_NAME` for `module_name` and `DEPENDENCY_NAME` for `dependency_name`?

guirodrigueslima: We don't need `module_name`, since the name used for the module inside` /opt/epics/module` is not relevant. What matters is allowing `dependency_name` to be defined, as files such as st.cmd and substitutions files may need to use this variable.

gustavosr8:
We should stick with the same style for the variables, IMO.
`REPONAME` and `MODULENAME` (or `MODNAME`?)
or
`REPO_NAME` and `MODULE_NAME`

guirodrigueslima:
> Na minha opinião, devemos manter o mesmo estilo para as variáveis. `REPONAME`e `MODULENAME`(ou `MODNAME`?) ou `REPO_NAME`e`MODULE_NAME`

As I understand it, the team is in favor of having "_" and changing REPONAME to REPO_NAME in the future. However, this change needs to be evaluated, as it will affect the build configuration of the current IOCs.

For this reason, I opted to keep MODULE_NAME.

PR #194, opened 2026-09-18
base: remove substitute command from pmac template

guirodrigueslima:
Remove the substitute command from the dls_pmac_asyn_motor template file because it depends on MSI for processing. Since MSI is not used in our build workflow, the command causes an error when starting the IOC.

# Checklist


gustavosr8:
@guirodrigueslima the MSI is part of the epics-base build system, which we are using to build the images. I think that I need to see the IOC logs to understand what is wrong about using the template, but it seems to me that it might be failing when looking for the `PMAC` and `ADDR` environment variables.

